PR #1, opened 2018-10-15
Bundle issues

artemfastdev:
Fix all import issues (sockets, stomps, history); add StaticRouter and BrowserRouter; add ports for local environment; fix package.json

### DIFF
--- a/src/api/apollo.ts
+++ b/src/api/apollo.ts
@@ -1,3 +1,4 @@
+import fetch from 'node-fetch'
 import ApolloClient from 'apollo-boost'
 
-export default new ApolloClient({ uri: '/api/graphql' })
+export default new ApolloClient({ uri: '/api/graphql', fetch: fetch })
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -2,13 +2,13 @@
 /* global Raven, process*/
 import { createStore, combineReducers, applyMiddleware } from 'redux';
 import { routerReducer, routerMiddleware } from 'react-router-redux';
-import createBrowserHistory from 'history/createBrowserHistory';
+import createMemoryHistory from 'history/createMemoryHistory';
 import { reducer as reduxFormReducer } from 'redux-form';
 import createSagaMiddleware from 'redux-saga';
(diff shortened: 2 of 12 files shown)


src/store/index.ts
```diff
@@ -2,13 +2,13 @@
...
 import { routerReducer, routerMiddleware } from 'react-router-redux';
-import createBrowserHistory from 'history/createBrowserHistory';
+import createMemoryHistory from 'history/createMemoryHistory';
 import { reducer as reduxFormReducer } from 'redux-form';
 import createSagaMiddleware from 'redux-saga';
 import rootSaga from './sagas';
 import reducers from './reducers';
 
-export const history = createBrowserHistory();
+export const history = createMemoryHistory();
```
hedvigoscar: This needs to be dependent on whether we are running the code in SSR mode or client mode. On client we need to use `createBrowserHistory()` and on server we need to use `createMemoryHistory()`

See https://github.com/ReactTraining/react-router/issues/4977#issuecomment-294394380 for more details 🙂

src/api/apollo.ts
```diff
@@ -1,3 +1,4 @@
+import fetch from 'node-fetch'
 import ApolloClient from 'apollo-boost'
 
-export default new ApolloClient({ uri: '/api/graphql' })
\ No newline at end of file
+export default new ApolloClient({ uri: '/api/graphql', fetch: fetch })
```
hedvigoscar: This needs to be dependent on whether we are running in SSR mode or not: see https://github.com/HedvigInsurance/web-onboarding/blob/master/src/clientEntry.tsx#L14 and https://github.com/HedvigInsurance/web-onboarding/blob/master/src/server/page.tsx#L65 for details.

hedvigoscar:
We also need to make sure that the code passes the CI build: right now there are issues caused by non-conformance to our linting guidelines. :+1:

hedvigoscar:
Seems like the build is still not working - but I think this is mostly fine. The old code that was not TypeScript does not need to pass Type checks - we can turn off strict type checking for these files.

PR #2, opened 2018-10-24
Fixes

artemfastdev:
Add and create proxy server to maven-project; change uri in apollo-client; add routes for koa-router


src/serverEntry.tsx
```diff
@@ -60,6 +68,14 @@ const server = createKoaServer({
 })
 
 server.router.get('/', getPage)
+reactPageRoutes.forEach((route) => {
+  server.router.get(route.path, getPage)
+})
+
+server.app.use(proxy({
+  target: 'http://localhost:8443',
```
hedvigoscar: I think this value should be configurable via environment variable.

```suggestion
  target: process.env.API_URL
```

PR #3, opened 2018-10-29
Use pagination and sorting of members on backend

artemfastdev:


hedvig.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   clientPath: path.resolve(__dirname, 'build/assets/'), // Build asset path
   serverPath: path.resolve(__dirname, 'build/'), // Build asset path
   port: 9083, // The WDS port
-  developmentPublicPath: "http://0.0.0.0:9083/", // Client public path during development, i.e. "http://0.0.0.0:8081/". Port must match the port directive
+  developmentPublicPath: "http://localhost:9083/", // Client public path during development, i.e. "http://0.0.0.0:8081/". Port must match the port directive
```
hedvigoscar: Is this change necessary? 🙂

hedvigoscar: Fair enough!

src/components/members/members-filter/MembersBackendFilter.tsx
```diff
@@ -0,0 +1,105 @@
...
+  members: MembersStore
+  search: (filter: MemberSearchFilter) => void
+  resetSearch: () => void
+}
+
+interface MembersBackendFilterState {
+  searchValue: string
+}
+
+export default class MembersBackendFilter extends React.Component<
```
Iamsamwhoami: We avoid using classes and instead mostly rely on stateless functional components. For state you could use https://github.com/diegohaz/constate and for side-effects we have our own library https://github.com/HedvigInsurance/react-lifecycle-components.

PR #6, opened 2018-10-30
Questions: load answered and not-answered separately

alex-bel:


src/components/questions/index.tsx
```diff
@@ -1,49 +1,48 @@
...
-  public tabChange = () => {
-    this.props.questionsRequest()
+const Questions = ({
+  questions,
+  members,
+  sendAnswer,
+  sendDoneMsg,
+  questionsRequest,
+}: QuestionsProps) => {
+  const tabChange = (e, data) => {
```
hedvigoscar: You can inline this as well - in general if functions are only used once it's most likely best to inline them 🙂

src/components/questions/questions-list/QuestionsList.tsx
```diff
@@ -17,11 +17,11 @@ const QuestionsList = ({
   sendAnswer,
   tabChange,
   sendDoneMsg,
-}) => {
+}: any) => {
```
hedvigoscar: Avoid `any` - try to use an interface and `React.SFC<T>` as often as possible

src/lib/helpers.ts
```diff
@@ -150,23 +151,20 @@ export const getClaimFieldsData = (
...
-export const sortQuestions = (questions) => ({
-  answered: questions.answered.sort((a, b) =>
-    moment(a.date).diff(moment(b.date)),
-  ),
-  notAnswered: questions.notAnswered.sort((a, b) =>
-    moment(a.date).diff(moment(b.date)),
-  ),
-})
+export const sortQuestions = (questions: QuestionGroup[]): QuestionGroup[] =>
+  questions.sort((a, b) => moment(a.date).diff(moment(b.date)))
```
hedvigoscar: Use `date-fns` instead of `moment`

PR #8, opened 2018-11-01
Fix/minor fixes

artemfastdev:
Minor fixes for chats, modals, tabs

### DIFF
--- a/src/clientEntry.tsx
+++ b/src/clientEntry.tsx
@@ -5,11 +5,14 @@ import { ApolloProvider } from 'react-apollo'
 import * as ReactDOM from 'react-dom'
 import { BrowserRouter } from 'react-router-dom'
 
+const appElement = document.getElementById('react-root');
+appElement.style.height = '100%';
+
 ReactDOM.hydrate(
   <BrowserRouter>
     <ApolloProvider client={apolloClient!}>
       <HotApp />
     </ApolloProvider>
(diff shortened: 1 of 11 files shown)


src/clientEntry.tsx
```diff
@@ -5,11 +5,14 @@ import { ApolloProvider } from 'react-apollo'
 import * as ReactDOM from 'react-dom'
 import { BrowserRouter } from 'react-router-dom'
 
+const appElement = document.getElementById('react-root')
+appElement.style.height = '100%'
```
hedvigoscar: I think this is more appropriate to put in an Emotion GlobalStyles-component - see https://github.com/HedvigInsurance/web-onboarding/blob/master/src/utils/globalStyles.tsx#L31 for inspiration 🙂

src/components/chat/tabs/InsuranceTab.tsx
```diff
@@ -129,11 +129,13 @@ export default class InsuranceTab extends React.Component {
               <Table.HeaderCell />
               <Table.HeaderCell colSpan="2">
                 <Modal
+                  style={{ height: window.innerHeight + 100 }}
```
hedvigoscar: I think we should use a styled component via `emotion` instead:

```js
const WideModal = styled(Modal)({
  height: window.innerHeight + 100
})
```

PR #11, opened 2018-11-06
Fix css-bug of date picker

artemfastdev:
Fix css-bug of date picker

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "nodemon": "^1.18.4",
     "react": "^16.5.2",
     "react-apollo": "^2.2.4",
-    "react-dates": "^18.1.0",
+    "react-dates": "^16.0.2",
     "react-dom": "^16.5.2",
     "react-emotion": "^9.2.10",
     "react-google-button": "^0.5.3",
--- a/src/components/shared/inputs/DateInput.tsx
+++ b/src/components/shared/inputs/DateInput.tsx
@@ -6,10 +6,11 @@ import * as React from 'react'
 import { SingleDatePicker } from 'react-dates'
 import { OPEN_UP } from 'react-dates/constants'
 import 'react-dates/initialize'
+import styled from 'react-emotion'
 import { Form } from 'semantic-ui-react'
-import styled from 'styled-components'
+import { styles } from './_dateinput'
 
-const WidgetContainer = styled.div`
+const WidgetContainer = styled('div')`
   display: flex;
   flex-direction: column;
   align-items: flex-start;
@@ -21,7 +22,7 @@ const WidgetContainer = styled.div`
   }
 `
 
-const DatePickerContainer = styled.div`
+const DatePickerContainer = styled('div')`
   font-family: 'Circular Std Book' !important;
   .SingleDatePickerInput {
     border: 1px solid #ccc;
@@ -69,7 +70,7 @@ export default class DateInput extends React.Component {
       <React.Fragment>
         <Form.Field disabled={this.props.disabled}>
           {this.props.label ? <label>Date</label> : null}
-          <WidgetContainer>
+          <WidgetContainer className={styles}>
             <DatePickerContainer>
               <SingleDatePicker
                 date={this.state.date}
--- a/src/components/shared/inputs/_dateinput.ts
+++ b/src/components/shared/inputs/_dateinput.ts
@@ -1,0 +1,825 @@
+import { css } from 'react-emotion'
+
+export const styles = css`
+  .PresetDateRangePicker_panel {
+    padding: 0 22px 11px;
+  }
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
     "nodemon": "^1.18.4",
     "react": "^16.5.2",
     "react-apollo": "^2.2.4",
-    "react-dates": "^18.1.0",
+    "react-dates": "^16.0.2",
```
hedvigoscar: Why have we gone from `18.1.0` to `16.0.2`?

artemfastdev: yep, seems like a newer version works a little bit strange with current settings (e.x. sometimes arrows are disappeared)

hedvigoscar: Ok! I think it would be preferable to make it work with the latest version instead.

src/components/shared/inputs/DateInput.tsx
```diff
@@ -40,61 +42,68 @@ const DatePickerContainer = styled.div`
...
+const DateInput: React.SFC<DateInputProps> = (props) => {
+  const initialState: State = {
+    focused: false,
   }
 
-  public componentWillMount() {
-    this.setState({ date: moment(this.props.date || moment()) }, () => {
-      this.dateChangeHandler(this.state.date)
-    })
+  const actions: ActionMap<State, Actions> = {
```
hedvigoscar: Move this to outside the render function body and pass in its dependencies where necessary. It does not need access to the entire closure

src/components/shared/inputs/DateInput.tsx
```diff
@@ -40,61 +42,68 @@ const DatePickerContainer = styled.div`
...
+    dateChangeHandler: (newDate: any) => (state) => {
+      props.changeHandler(props.changeType || DATE, null, {
+        value: moment(newDate).toISOString(),
```
hedvigoscar: Use `date-fns` here 🙂

src/components/shared/inputs/DateInput.tsx
```diff
@@ -40,61 +42,68 @@ const DatePickerContainer = styled.div`
...
+      })
+      return { date: moment(newDate) }
```
hedvigoscar: same here

src/components/shared/inputs/DateInput.tsx
```diff
@@ -1,15 +1,17 @@
...
 import { OPEN_UP } from 'react-dates/constants'
 import 'react-dates/initialize'
+import styled from 'react-emotion'
+import { Mount } from 'react-lifecycle-components'
 import { Form } from 'semantic-ui-react'
-import styled from 'styled-components'
+import { dateInputStyles } from './_dateinput'
 
-const WidgetContainer = styled.div`
+const WidgetContainer = styled('div')`
```
hedvigoscar: We prefer the object-based syntax: 
```suggestion
const WidgetContainer = styled('div')({
  display: 'flex',
```

src/components/shared/inputs/DateInput.tsx
artemfastdev: sorry, Oscar, but it'll break current date picker. And if i try to use material-ui one, it requires to use alpha-version of date-fns and that's not very cool for entire settings

src/components/shared/inputs/DateInput.tsx
artemfastdev: okay, np

src/components/shared/inputs/DateInput.tsx
artemfastdev: no problem

src/components/shared/inputs/DateInput.tsx
hedvigoscar: But all you need to do is convert this date to an iso-string, correct?

Then you should be able to do:

```ts
import parse from 'date-fns/parse'

...
const parsed = parse(date) // => returns a Date()
const isoString = parsed.toISOString() // => converts date to ISO string by regular means
```

src/components/shared/inputs/DateInput.tsx
hedvigoscar: if `react-dates` uses moment then by all means you may continue to use it, but otherwise try to use `date-fns` as much as possible 🙂

src/components/shared/inputs/DateInput.tsx
artemfastdev: i suggest you to leave as it is, for now and create a new task in trello about finding the right solution for the situation. What do you think?

hedvigoscar: Given `react-dates` dependency on `moment` I guess we can skip it for now.

As a rule we should try to improve things as part of the tickets in progress though 🙂

PR #18, opened 2018-11-16
Indicator of fraudulent status of the member

dkaplunov-fd:
... and trace who change the member entity


src/components/chat/tabs/insurance-trace/InsuranceTrace.tsx
```diff
@@ -2,13 +2,15 @@ import { getFieldName } from 'lib/helpers'
 import * as React from 'react'
 import { Table } from 'semantic-ui-react'
 
+const TextWrapper = ({ text, width }) => (
+  <div style={{ wordWrap: 'break-word', width: width || '150px' }}>{text}</div>
```
palmfjord: no inline styling plz. you can make a styled component instead like so:
```tsx
const TextWrapper = styled('div')(({ width = 150 }: { width?: number }) => ({
  wordWrap: 'break-word',
  width,
}))

// ...

<TextWrapper width={123}>the text</TextWrapper>
```

dkaplunov-fd: Done

src/components/chat/tabs/insurance-trace/InsuranceTrace.tsx
```diff
@@ -1,9 +1,16 @@
 import { getFieldName } from 'lib/helpers'
 import * as React from 'react'
 import { Table } from 'semantic-ui-react'
+import styled from 'styled-components'
 
-const TextWrapper = ({ text, width }) => (
-  <div style={{ wordWrap: 'break-word', width: width || '150px' }}>{text}</div>
+const TextWrapper = styled('div')(
+  (
+    { width = '200px' }: { width?: string },
```
palmfjord: nice one!

src/components/chat/tabs/DetailsTab.tsx
```diff
@@ -50,73 +55,105 @@ export default class DetailsTab extends React.Component {
...
+        <Table selectable>
+          <Table.Body>
+            <TableFields fields={memberInfo} />
+            <FraudulentStatusEdit
+              getFraudStatusInfo={() => ({
+                status: this.state.fraudStatus || fraudulentStatus,
+                description:
+                  this.state.fraudDescription || fraudulentDescription,
+              })}
+              setState={(val, fs, desc) => {
```
palmfjord: `fs` is kinda claimed by 'file system', just saying :D

src/components/chat/tabs/DetailsTab.tsx
```diff
@@ -50,73 +55,105 @@ export default class DetailsTab extends React.Component {
...
+                  size="small"
+                  dimmer="blurring"
+                >
+                  <Header icon="edit" content="Edit Member" />
+                  <Modal.Content>
+                    <Form inverted size="small">
+                      <React.Fragment>
+                        {Object.keys(memberInfo).map((field, id) => (
+                          <Form.Input
+                            key={id}
```
palmfjord: it's not so good to use the array index as they `key` as it will always be ordinal, so if this list changes for some reason we might end up with the wrong fields in the DOM. I'm pretty sure we could use the `field` value instead, as an object key should never occur twice

src/components/chat/tabs/DetailsTab.tsx
```diff
@@ -50,73 +55,109 @@ export default class DetailsTab extends React.Component {
...
+                  open={this.state.modalOpen}
+                  onClose={this.handleClose}
+                  basic
+                  size="small"
+                  dimmer="blurring"
+                >
+                  <Header icon="edit" content="Edit Member" />
+                  <Modal.Content>
+                    <Form inverted size="small">
+                      <React.Fragment>
```
palmfjord: `<></>` :)

src/components/chat/tabs/DetailsTab.tsx
```diff
@@ -126,4 +167,5 @@ export default class DetailsTab extends React.Component {
 DetailsTab.propTypes = {
   messages: PropTypes.object.isRequired,
   editMemberDetails: PropTypes.func.isRequired,
+  saveFraudulentStatus: PropTypes.func.isRequired,
```
palmfjord: since this is typescript nowadays we should kill the prop types in favour of types, but maybe we can leave it for later

src/lib/fraudulentStatus.tsx
```diff
@@ -0,0 +1,98 @@
...
+            />
+            <br />
+            <Input
+              fluid
+              onChange={(e, d) => (descriptionValue = d.value)}
+              defaultValue={descriptionValue}
+            />
+          </>
+        )}
+        <div style={{ float: 'right', marginTop: '10px' }}>
```
palmfjord: no inline stylezzz

palmfjord: check this whole file, there's a few places

src/components/chat/tabs/DetailsTab.tsx
dkaplunov-fd: done

src/lib/fraudulentStatus.tsx
dkaplunov-fd: Done, I hope...

PR #21, opened 2018-11-22
Emoji picker :tada:

hedvigoscar:


src/components/chat/chat/EmojiPicker.tsx
```diff
@@ -0,0 +1,52 @@
...
+interface State {
+  open: boolean
+}
+
+interface Actions {
+  setIsOpen: (open: boolean) => void
+}
+
+const actions: ActionMap<State, Actions> = {
+  setIsOpen: (open: boolean) => (state) => ({ ...state, open }),
```
palmfjord: i think `constate` merges the state for us, so technically we could just do `({ open })`

hedvigoscar: I think you're right, this is probably cruft.

```suggestion
  setIsOpen: (open: boolean) => () => ({ open }),
```

PR #27, opened 2018-12-04
Change format of "List of change"

dkaplunov-fd:


src/components/chat/tabs/insurance-trace/InsuranceTrace.tsx
```diff
@@ -8,25 +8,30 @@ const TextWrapper = styled('div')(
...
+  color: '#888888',
+})
+
 const InsuranceTrace = ({ traceData }) =>
   !traceData || traceData.length === 0 ? null : (
     <>
       <h4>List of changes</h4>
       <ScrollList selection>
-        {traceData.map((trace) => (
+        {traceData.reverse().map((trace) => (
```
palmfjord: this'll mutate `traceData`, maybe we should do `[...traceData].reverse()` just to be on the safe side?

dkaplunov-fd: done

PR #33, opened 2018-12-07
Add deductible field to frontend

dkaplunov-fd:


palmfjord:
idk what this is but i will close it now

PR #34, opened 2018-12-18
Fix payment context bug with nested Formik forms

isaacSennerholt:


src/components/claims/claim-details/components/PaymentConfirmationDialog.tsx
```diff
@@ -98,13 +101,14 @@ export const PaymentConfirmationDialog: React.SFC<
                   color="primary"
                   disabled={!isValid}
                   variant="contained"
+                  onClick={() => submitForm()}
```
palmfjord: it's weird that this doesn't work tho - since the button has type `submit`. U sure this is just not masking a different problem? Just checking :)

palmfjord: especially since i would assume this works? https://github.com/HedvigInsurance/hedgehog/pull/34/files#diff-331b1b321642c7c5c24826972797919bR371

palmfjord: also - u sure it wont submit this twice?

isaacSennerholt: The behavior was that the nested form hijacked the context of the parent form. I don't know exactly why that is, but it did. The result was that the nested form submitted itself and the parent form on submit. It might be due to a name space conflict or something of sort. Therefor i solved it by utilising explicit closures.

palmfjord: ugh, that kinda makes sense actually. Kinda. Anyways forms shouldn't really be nested, like... ever. But if this solves it lets leave it for now. However - what happens if you press `enter` or something when focusing on one of the fields? Won't that trigger the exact same behavior?

palmfjord: nice catch tho! Sounds like something that would be hard to track down?

isaacSennerholt: Damn, you're right! Didn't catch the onEnter issue while testing. This fix did not resolve that issue. I guess i have to fall back on the solution of moving the nested form out of the parent form.. Though i didn't figure out a good way of accessing the "resetForm" of the parent form in a "not to messy" way.. Any ideas?

isaacSennerholt: Lol, i forgot that Formik is just a state container and not the actual form.. I've created a new PR with a very obvious solution.

PR #43, opened 2019-01-04
Link to member page from claims view

hedvigoscar:


src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -70,6 +73,9 @@ const MemberInformation: React.SFC<MemberInformationProps> = ({
 }) => (
   <Paper>
     <h3>Member Information</h3>
+    <p>
+      Id: <Link to={`/members/${memberId}`}>{memberId}</Link>
+    </p>
```
isaacSennerholt: Can we move this beneath MemberName? :)

hedvigoscar: Done!

src/components/claims/claim-details/index.tsx
```diff
@@ -14,6 +14,7 @@ const CLAIM_PAGE_QUERY = gql`
   query ClaimPage($id: ID!) {
     claim(id: $id) {
       member {
+        memberId
```
palmfjord: maybe we should add this id to `dataIdFromObject` to the cache? So it normalizes correctly https://www.apollographql.com/docs/react/advanced/caching.html#normalization

PR #47, opened 2019-01-09
Add date to claim notes

isaacSennerholt:


src/components/claims/claim-details/components/ClaimNotes.tsx
```diff
@@ -65,12 +71,30 @@ const TextArea: React.SFC<FieldProps<HTMLTextAreaElement>> = ({
   />
 )
 
+const sortNotesByDate = (notes: Note[]) =>
+  notes.sort((noteA, noteB) => {
```
palmfjord: `sort` mutates the array, so we should probably do something like `[...notes].sort` to be sure we don't accidentally sort something we don't want sorted

isaacSennerholt: Good catch!

PR #56, opened 2019-04-09
Dev experience updates

palmfjord:
makes life slightly easier (i hope <- tihi)


src/lib/checkAuth.ts
```diff
@@ -11,6 +12,21 @@ export function checkLocalAuth(getState) {
...
+    return await api(loginConfig)
+  } catch (e) {
+    const isNetworkError = (e.message as string).includes('Network Error')
+    if (process.env.NODE_ENV !== 'development' || !isNetworkError) {
+      throw e
+    }
+
+    await new Promise((resolve) => setTimeout(resolve, 1000))
+
+    return api(config.login.login)
```
meletios: is it `loginConfig` here?

palmfjord: no 😬 but this line is correct. resolved after irl discussion

PR #60, opened 2019-05-10
Add logging to places where redirect to /login/oauth

johantjellden:


src/lib/checkAuth.ts
```diff
@@ -40,6 +41,7 @@ export async function checkApiAuth({ dispatch, getState }) {
     dispatch(setClient(client.data))
     return true
   } catch (error) {
+    Sentry.captureMessage(`Logout from checkApiAuth in lib/checkAuth.ts/checkApiAuth() with error: ${error.message}`);
```
palmfjord: maybe we should explicitly set the log level to `info`? with
```ts
Sentry.captureMessage(`Logout from checkApiAuth in lib/checkAuth.ts/checkApiAuth() with error: ${error.message}`, { level: 'info' })
```

also no semis plz 😬

package.json
```diff
@@ -25,6 +25,7 @@
     "@hedviginsurance/web-survival-kit": "^1.8.2",
     "@material-ui/core": "^3.5.1",
     "@material-ui/icons": "^3.0.1",
+    "@sentry/browser": "5.2.0",
```
palmfjord: this should also be in the `yarn.lock` or it wont install properly on heroku, use `yarn add @sentry/browser` and `yarn add dotenv` when you install packages

example.env
```diff
@@ -0,0 +1 @@
+SENTRY_DSN=https://XXXX
```
palmfjord: imo this should be `.env.example`,  it's just a convention :)

src/store/sagas/loginSaga.ts
```diff
@@ -4,6 +4,8 @@ import { call, cancel, cancelled, fork, put, take } from 'redux-saga/effects'
 import { history } from 'store'
 import { setClient, unsetClient } from '../actions/clientActions'
 import { showNotification } from '../actions/notificationsActions'
+import * as Sentry from '@sentry/browser'
```
palmfjord: this is super nitty-gritty but i think this should be moved up so the import is grouped with the rest of the imports from node_modules :)

package.json
palmfjord: and ofc make sure you check in the changes in yarn.lock :)

src/serverEntry.tsx
```diff
@@ -31,12 +34,19 @@ const template = () => `
...
-          crossorigin="anonymous"></script>
+  <script src="https://browser.sentry-cdn.com/5.2.0/bundle.min.js" crossorigin="anonymous"></script>
+
 </head>
 <body>
   <div id="react-root"></div>
 
+  <script>
+    Sentry.init({
+      dsn: '${process.env.SENTRY_DSN}',
```
palmenhq: We should probably use `${JSON.stringify(process.env.SENTRY_DSN)}` instead of wrapping it in quotes, much safer to prevent potential xss

src/serverEntry.tsx
```diff
@@ -31,12 +34,19 @@ const template = () => `
...
+      environment: '${process.env.NODE_ENV}'
```
palmenhq: this will always be `production` in heroku. Could we set a `SENTRY_ENVIRONMENT` env var instead? That way we can differentiate between staging and prod for example

johantjellden:
Lets close this.

PR #61, opened 2019-05-13
Feature/ fileupload table rendered in new tab within Member section

AlexSJ1990:


src/components/chat/tabs/FileTab.tsx
```diff
@@ -0,0 +1,92 @@
...
+      fileUploads {
+        fileUploadUrl
+        memberId
+        timestamp
+        mimeType
+      }
+    }
+  }
+`
+const fileDateSorter = (a, b) => {
```
palmfjord: Imo functions should be named as verbs, so `sortFileDate` would be a more appropriate name :)

palmfjord: also, actually i _think_  this could be simplified to `(a, b) => a -  b` (slightly unsure i think i remember correctly)

src/components/chat/tabs/FileTab.tsx
```diff
@@ -0,0 +1,92 @@
...
+    if (aDate > bDate) {
+      return -1
+    }
+    if (bDate > aDate) {
+      return 1
+    }
+    return 0
+}
+
+const TableExampleCelled = ( { memberFiles } ) => (
```
palmfjord: no spaces between `( {` :D you can install prettier for your editor so you don't have to think about this

src/components/chat/tabs/FileTab.tsx
```diff
@@ -0,0 +1,92 @@
...
+                <Table.HeaderCell>Member File</Table.HeaderCell>
+                <Table.HeaderCell>Time Stamp</Table.HeaderCell>
+                <Table.HeaderCell>File Type</Table.HeaderCell>
+            </Table.Row>
+        </Table.Header>
+        <Table.Body>
+            { memberFiles.length === 0 
+            ? "No files uploaded for this member yet" 
+            : memberFiles.sort(fileDateSorter).map((memberFile) => (
+                <Table.Row>
```
palmfjord: this must have a `key={...}` prop or react will be sad

src/components/chat/tabs/FileTab.tsx
```diff
@@ -0,0 +1,92 @@
...
+                    <Table.Cell>
+                        <Image src={ memberFile.fileUploadUrl } size="medium" />
```
palmfjord: formatting is a bit off here too but it should fix itself with prettier :)

src/components/chat/tabs/FileTab.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <Table.Header>
+            <Table.Row>
+                <Table.HeaderCell>Member File</Table.HeaderCell>
+                <Table.HeaderCell>Time Stamp</Table.HeaderCell>
+                <Table.HeaderCell>File Type</Table.HeaderCell>
+            </Table.Row>
+        </Table.Header>
+        <Table.Body>
+            { memberFiles.length === 0 
+            ? "No files uploaded for this member yet"
```
palmfjord: just asking, should this really be in the table body? Isn't it better to either show this _or_ the table?

src/components/chat/tabs/FileTab.tsx
```diff
@@ -0,0 +1,92 @@
...
+                </Table.Row>
+            ))}
+        </Table.Body>
+    </Table>
+)
+
+class MemberFile extends React.Component {
+    constructor(props) {
+        super(props)
+        this.state = { clicked: false }
```
palmfjord: we usually initialize this outside the constructor, like so:
```js
class MemberFile extends React.Component {
  state = { clicked: false }
...
}
```

src/components/chat/tabs/FileTab.tsx
```diff
@@ -0,0 +1,92 @@
...
+        this.variables = {
```
palmfjord: why do we need to assign this? It feels shaky as it won't update if the component gets a new `props.match.params.id`

src/components/chat/tabs/FileTab.tsx
palmfjord: also maybe we should name this component something more meaningful

src/components/chat/tabs/FileTab.tsx
```diff
@@ -0,0 +1,92 @@
+import gql from "graphql-tag";
+import React from "react";
```
palmfjord: does this work? As it's typescript and not javascript this should probably be `import * as React from 'react'`, but maybe this particular repository has the compat mode on, i'm unsure

PR #63, opened 2019-05-23
Feature/payment schedule graphql

AlexSJ1990:


src/features/charges/index.tsx
```diff
@@ -33,71 +63,157 @@ const Button = styled('button')({
...
+                                        (payment) => ({
+                                          memberId: payment.member.memberId,
+                                          amount:
+                                            payment.member.account
+                                              .currentBalance,
+                                        }),
+                                      ),
+                                    },
+                                  }).then(console.log)
+                                  this.resetButton()
```
palmenhq: this needs to happen once the mutation is done

src/features/charges/index.tsx
```diff
@@ -120,15 +122,23 @@ interface State {
...
     confirming: false,
   }
 
   public render() {
     return (
       <Wrapper>
-        <Query query={query} variables={{ month: moment().format('YYYY-MM') }}>
+        <Query<any>
+          query={query}
+          variables={{ month: moment().format('YYYY-MM') }}
```
palmenhq: Ah! We should probably use `date-fns` instead of `moment` (as we're trying to phase out the latter)

PR #64, opened 2019-05-28
Feature/automatic payments

vonElfvin:


src/components/chat/tabs/account-tab/add-entry-form.tsx
```diff
@@ -0,0 +1,266 @@
...
+  },
+})
+
+const parseAmount = (amount: string) => parseFloat(amount.replace(/\s+/g, ''))
+
+const getValidationSchema = () =>
+  yup.object().shape({
+    type: yup
+      .string()
+      .oneOf(['CORRECTION', 'CAMPAIGN', 'SUBSCRIPTION', 'PAYOUT', 'CHARGE'])
```
palmenhq: This is wrong now, we should fix this

PR #66, opened 2019-07-01
Auto answer suggestion

hkindbom:
To review before pushing to production


src/components/chat/tabs/ChatTab.tsx
```diff
@@ -51,7 +64,38 @@ export default class ChatTab extends React.Component {
...
 
+  //concatenates last messages written by member into one question. The loop could maybe be improved?
+  private getQuestionToAnalyze(){
+
+    var lastMemberMessages = '';
+    var messageIds = [];
+
+    if (this.props.messages && this.props.messages.list){
+            
+      let message = this.props.messages.list[this.props.messages.list.length-1];
```
palmenhq: this'll probably explode if the message list is empty

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -51,7 +64,38 @@ export default class ChatTab extends React.Component {
     }
   }
 
+  //concatenates last messages written by member into one question. The loop could maybe be improved?
+  private getQuestionToAnalyze(){
+
+    var lastMemberMessages = '';
+    var messageIds = [];
+
+    if (this.props.messages && this.props.messages.list){
```
palmenhq: instead of wrapping everything in `if`s we could make some early returns :) so we invert these bool expressions and return inside the if instead. Something like

```js
if (!this.props.messages || !this.props.messages.list) {
  return [lastMemberMessages, messageIds]
}

let i = 12
// ...
```

imo this strategy makes the code way easier to read since the indentation level doesn't stack up and we only have one `if` at a time to keep track of

palmenhq: this applies to a few places actually

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -51,7 +64,38 @@ export default class ChatTab extends React.Component {
...
+            
+      let message = this.props.messages.list[this.props.messages.list.length-1];      
+
+      let i=1;
+      // this.props.match.params.id is the member id
+      while (i<this.props.messages.list.length+1 && message.header.fromId == +this.props.match.params.id && message.id === "free.chat.message"){
```
palmenhq: prefer ie [`Array#map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) and things like that over loops, much cleaner

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -67,11 +111,36 @@ export default class ChatTab extends React.Component {
             messageId={
               (this.props.match && this.props.match.params.msgId) || ''
             }
-          />
-          <ChatPanel
+          />    
+
+          {/* alternatives for updating the query https://www.apollographql.com/docs/react/essentials/queries/#polling-and-refetching*/}
```
palmenhq: i think these types of comments in general are a bit redundant and may get outdated quickly. Generally it's quite easy to find documentation yourself

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -67,11 +111,36 @@ export default class ChatTab extends React.Component {
...
+          <Query query={GET_SUGGESTED_ANSWER_QUERY} pollInterval={2000} variables={{question: questionAndMessageIds[0]}}>
+          {({ data, loading, error }) => {
+            if (loading || error) return <ChatPanel
```
palmenhq: plz use brackets (`if (...) { ... }`), it's too easy to introduce fall-through statements if we don't wrap things in brackets

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -67,11 +111,36 @@ export default class ChatTab extends React.Component {
...
+            questionToLabel = ''
+            addMessage={this.props.addMessage}
+            messages={(this.props.messages && this.props.messages.list) || []}
+            suggestedAnswer = ''
+          />;
+    
+            return (
+
+            <ChatPanel
+            allReplies = {JSON.parse(data.getAnswerSuggestion.allReplies)}
```
palmenhq: why is `data.getAnswerSuggestion.allReplies` a string? seems smelly to me

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -103,21 +132,73 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
+                {this.props.questionToLabel !== '' && (            
+              
+                <div>
+                <MenuList>
+
+                {allIntents!.map((intent) => {
+                  //strip \n 
+                  const  text  = allReplies[intent].reply.replace(/(\r\n|\n|\r)/gm, "");
+                  return (
+                     (this.state.showMoreReplies == true || (this.state.showMoreReplies == false && this.props.suggestedAnswer === text)) && (
```
palmenhq: `===` > `==`. Also there's no need to compare a boolean to a boolean, `this.state.showMoreReplies || (!this.state.showMoreReplies && this.props.suggestedAnswer === text)` should be sufficient and is fewer keystrokes 😉

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -103,21 +132,73 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
-            variant="outlined"
-          />
+
+    const allReplies = this.props.allReplies;
+    const allIntents = this.getAllIntents(allReplies);
+
+
+    return (    
+   
+          <Mutation mutation= {AUTO_LABEL_QUESTION} ignoreResults = {true} onCompleted = {this.sendMessage}
```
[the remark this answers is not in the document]
hkindbom: What do you mean exactly?

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -103,21 +132,73 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
+                <div>
+                <MenuList>
+
+                {allIntents!.map((intent) => {
+                  //strip \n 
+                  const  text  = allReplies[intent].reply.replace(/(\r\n|\n|\r)/gm, "");
+                  return (
+                     (this.state.showMoreReplies == true || (this.state.showMoreReplies == false && this.props.suggestedAnswer === text)) && (
+                      <MenuItem
+                      key={text}
```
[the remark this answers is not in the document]
hkindbom: You mean since the key is "text"? I don't think we will ever have two identical texts but how can i solve it easily anyways?

src/components/chat/tabs/ChatTab.tsx
hkindbom: I thought it was a quick and simple solution for now. Didn't know how to handle it in a good way in backoffice

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -217,12 +317,43 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
       })
   }
 
+  //takes an object with intents as keys
+  private getAllIntents = (allReplies: array) => {
+
+    var allIntents = [];
+
+    //appending all intents/keys to array, except from other class
+    for (var key in allReplies) {
```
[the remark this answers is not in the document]
hkindbom: Why do we need to make a "Set"?

src/components/chat/tabs/ChatTab.tsx
hkindbom: changed to this.props.messages.list > 0, might work?

hkindbom: if (!this.props.messages || !this.props.messages.list || this.props.messages.list.length === 0)){
      return [lastMemberMessages, messageIds];} should do it?

src/components/chat/chat/ChatPanel.tsx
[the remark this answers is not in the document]
palmenhq: you can run `yarn prettier --write` i think and it should format the code with the rest of the code base

src/components/chat/chat/ChatPanel.tsx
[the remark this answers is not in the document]
palmenhq: if there's an id or something you should use that. If no then i guess this is _ok_

src/components/chat/tabs/ChatTab.tsx
palmenhq: Preferably you would let back-office parse it

src/components/chat/chat/ChatPanel.tsx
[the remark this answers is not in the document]
palmenhq: `Set` will remove all duplicate values, that's kind of what you're doing right?

src/components/chat/tabs/ChatTab.tsx
palmenhq: looks good 👍

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -103,80 +141,162 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
+                <MuiIcon>send</MuiIcon>
+              </SubmitButton>
+            </OptionsContainer>
+          </MessagesPanelContainer>
+        )}
+      </Mutation>
+    )
+  }
+  private getReply = (allReplies: object, intent: string) => {
+    //strip \n
```
palmenhq: redundant comment :) i can read that this is happening from the code right below it. Also it's not completely true because it also strips `\r`

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -103,80 +141,162 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
+                Send
+                <MuiIcon>send</MuiIcon>
+              </SubmitButton>
+            </OptionsContainer>
+          </MessagesPanelContainer>
+        )}
+      </Mutation>
+    )
+  }
+  private getReply = (allReplies: object, intent: string) => {
```
palmenhq: Maybe this should be something like `getReplyOnOneLine` or something?

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -217,12 +341,33 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
   }
 
+  private getAllIntents = (allReplies: object) => {
+    const allIntents = allReplies
+      ? Object.keys(allReplies).filter((key) => key !== 'other')
+      : []
+    return [...new Set(allIntents)]
+  }
+
+  private showMoreReplies = () => {
```
palmenhq: technically shouldn't this be `toggleMoreReplies`?

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -51,7 +62,43 @@ export default class ChatTab extends React.Component {
     }
   }
 
+  //concatenates last messages written by member into one question. The loop could maybe be improved?
+  private getQuestionToAnalyze() {
+    var lastMemberMessages = ''
```
palmenhq: avoid `var` plz :)

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -51,7 +62,43 @@ export default class ChatTab extends React.Component {
     }
   }
 
+  //concatenates last messages written by member into one question. The loop could maybe be improved?
```
palmenhq: This comment is actually already outdated a bit (there's no loop anymore). If the method needs further explanation we should probably improve naming instead :)

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -51,7 +62,43 @@ export default class ChatTab extends React.Component {
...
+
+    const messages = this.props.messages.list
+
+    const fromIds = messages.map((message) => message.header.fromId)
+
+    const lastNonMemberIndex = fromIds
+      .map((id) => id === +this.props.match.params.id)
+      .lastIndexOf(false)
+
+    lastMemberMessages = messages.filter(
```
palmenhq: i think we should avoid these type of reassignments. Either we should chain the map/filter/whatever or we should make new variables so we don't accidentally mix things

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -51,7 +62,43 @@ export default class ChatTab extends React.Component {
...
+  private getQuestionToAnalyze() {
+    var lastMemberMessages = ''
+    var messageIds = []
+
+    if (
+      !this.props.messages ||
+      !this.props.messages.list ||
+      this.props.messages.list.length === 0
+    ) {
+      return [lastMemberMessages, messageIds]
```
palmenhq: shouldn't this be `{ lastMemberMessages, messageIds }` now?

src/components/chat/tabs/ChatTab.tsx
palmenhq: also why is this a string? it's getting transformed into an array later.

palmenhq: ah. It's getting transformed back into a string after that. i follow now. Feels a bit confusing though, i think we should avoid reusing variables like this.

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -103,80 +141,162 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
+        {(autoLabelQuestion) => (
+          <MessagesPanelContainer>
+            <ChatForm onSubmit={this.handleSubmit}>
+              {this.props.questionToLabel !== '' && (
+                <div>
+                  <MenuList>
+                    {allIntents!.map((intent) => {
+                      const text = this.getReply(allReplies, intent)
+                      return (
+                        (this.state.showMoreReplies ||
```
palmenhq: could we maybe refactor this into a named function somehow? Get's a bit hard to read with so many nested bool expressions

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -103,80 +141,162 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
+
+  private shouldSubmit = (e: React.KeyboardEvent<any>) => {
+    return (
+      !window.matchMedia('(max-width: 800px)').matches &&
+      e.keyCode === 13 &&
+      !e.shiftKey
     )
   }
 
   private handleEnterMaybe = (e: React.KeyboardEvent<any>) => {
```
palmenhq: this is now unused and can be removed no?

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -51,7 +65,60 @@ export default class ChatTab extends React.Component {
...
+      .map((message) => message.body.text)
+      .join(' ')
+
+    return { lastMemberMessages, messageIds }
+  }
+
+  private getQuestionAndAnswer(responses: object) {
+    let question = "";
+    let answer = "";
+    const questionResponses = responses.filter(sentence =>  (sentence.text.match(/\?/g) || []).length > 0);
```
palmenhq: optional change but could also be rewritten to something like `responses.filter(sentence => sentence.includes('?'))` to save a few keystrokes :)

src/components/chat/tabs/ChatTab.tsx
```diff
@@ -51,7 +65,60 @@ export default class ChatTab extends React.Component {
...
+
+    if (questionResponses.length > 1){
+      return {question, answer}
+    }else if(questionResponses.length === 1){
```
palmenhq: this formatting is off, you should run prettier on this file. My suggestion would be to do it often, or to use something like an [editor plugin](https://prettier.io/docs/en/editors.html) :)

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -13,6 +13,9 @@ import { format } from 'date-fns'
 import * as React from 'react'
 import styled from 'react-emotion'
 import { EmojiPicker } from './EmojiPicker'
+import axios from 'axios'
```
palmenhq: unused no?

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -103,80 +140,164 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
+                    autoLabelQuestion({
+                      variables: {
+                        question: this.props.questionToLabel,
+                        label: this.state.chosenIntent,
+                        memberId: this.props.memberId,
+                        messageIds: this.props.messageIds,
+                      },
+                    })
+                  } else {
+                    this.sendMessage()
```
palmenhq: i like this, i think we should do this also  on line 211 instead of doing it after the graphql query completes

src/components/chat/chat/ChatPanel.tsx
```diff
@@ -103,80 +140,161 @@ export class ChatPanel extends React.PureComponent<ChatPanelProps, State> {
...
-        <OptionsContainer>
-          <MuiFormControlLabel
-            label="Force message"
-            labelPlacement="start"
-            control={
-              <OptionCheckbox
-                color="primary"
-                checked={this.state.forceSendMessage}
-                onChange={this.handleCheckboxChange}
+      <Mutation mutation={AUTO_LABEL_QUESTION} ignoreResults={true}>
```
palmenhq: ooh nice! Didn't know `ignoreResults` existed

PR #67, opened 2019-07-18
Item pricing

RasmusHedvig:
Merge to master or die trying


src/components/pricing/FilterContainer.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <PropContainer>
+        {this.props.filters.length !== 0 ? (
+          <SubHeaderContainer>
+            <Header size="small">Active Filters</Header>
+          </SubHeaderContainer>
+        ) : (
+          <SubHeaderContainer />
+        )}
+        {this.props.filters.map((filterRow) => (
+          <Label
```
palmenhq: There should be a `key` prop on every component that is in an array, see https://reactjs.org/docs/lists-and-keys.html#keys

src/components/pricing/FilterContainer.tsx
```diff
@@ -0,0 +1,70 @@
...
+        {this.props.items.suggestions.length !== 0 ? (
+          <SubHeaderContainer>
+            <Header size="small">Suggested Filters</Header>
+          </SubHeaderContainer>
+        ) : (
+          <SubHeaderContainer />
+        )}
+
+        {this.props.items.suggestions.map((categoryRow) => (
+          <Label.Group color="blue">
```
palmenhq: key :)

src/components/pricing/FilterContainer.tsx
```diff
@@ -0,0 +1,70 @@
...
+            {categoryRow.items.map((itemRow) => (
+              <Label
```
palmenhq: aaand key

src/components/pricing/ItemTable.tsx
```diff
@@ -0,0 +1,83 @@
+import { GET_PRICES } from 'features/pricing/queries'
+import * as React from 'react'
+import { Query } from 'react-apollo'
+import { Label, Table } from 'semantic-ui-react'
+
+export default class ItemTable extends React.Component {
+  public handlePriceData = (data) => {
+    if (data) {
+      return 'prices' in data
+        ? Object.assign({}, ...data.prices.map((s) => ({ [s.itemId]: s })))
```
palmenhq: this is cleaver, however i think a more readable solution would be something like
```js
data.prices.reduce((accumulator, s) => ({ ...accumulator, [s.itemId]: s }), {})
```

src/components/pricing/ItemTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+        : {}
+    }
+
+    return {}
+  }
+
+  public getItemIds = () => {
+    const { products } = this.props.items
+    return [...new Set(products.slice(0, 10).map((item) => item.id))]
```
palmenhq: very clever!

src/components/pricing/ItemTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  public getPriceString = (prices, row, property) => {
+    return row.id in prices
+      ? Math.floor(prices[row.id][property]).toLocaleString('sv-SE') + ' kr'
```
palmenhq: actually we have a util function for formatting money already [here](https://github.com/HedvigInsurance/hedgehog/blob/5058d707d94594186e915f851a895b9b88a59683/src/lib/intl.ts#L12) and i think we should utilize it

src/components/pricing/ItemTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+        </Table.Header>
+
+        <Table.Body>
+          <Query
+            query={GET_PRICES}
+            variables={{ date: this.props.date, ids: this.getItemIds() }}
+          >
+            {({ loading, data }) => {
+              const prices = this.handlePriceData(data)
+              return this.props.items.products.slice(0, 10).map((row) => (
```
palmenhq: 🗝

src/components/pricing/ItemTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+                <Table.Row>
+                  <Table.Cell>{row.name}</Table.Cell>
+                  <Table.Cell textAlign="center" verticalAlign="middle">
+                    <Label
+                      basic
+                      as="a"
+                      color={
+                        loading ? 'grey' : row.id in prices ? 'green' : 'grey'
```
palmenhq: nested ternaries = hard to read :/ i think we should find a cleaner way to do this or just move it into a pure function with `if`s

src/components/pricing/ItemTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+                    >
+                      {loading ? '…' : this.getPriceString(prices, row, 'mean')}
+                    </Label>
+                  </Table.Cell>
+                  <Table.Cell textAlign="center" verticalAlign="middle">
+                    <Label
+                      basic
+                      as="a"
+                      color={
+                        loading ? 'grey' : row.id in prices ? 'blue' : 'grey'
```
palmenhq: same here, avoid nested ternaries plz :)

src/features/pricing/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+        activeFilters: [
+          ...state.activeFilters,
+          { value: itemRow, name: categoryRow.name },
+        ],
+      }
+    })
+  }
+
+  public removeFilter = (event, filterRow) => {
+    this.setState((state) => {
```
palmenhq: beautiful 😍

src/features/pricing/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+      return {
+        activeFilters: state.activeFilters.filter(
+          (e) => e.name !== filterRow.name,
+        ),
+      }
+    })
+  }
+
+  public isDateValid = (attemptedDate) => {
+    const date = Date.parse(attemptedDate)
```
palmenhq: this is clever, but also very fragile tbh 😬 maybe we could utilize `date-fns`'s [`isValid`](https://date-fns.org/docs/isValid) n stuff instead?

src/components/pricing/ItemTable.tsx
palmenhq: although, maybe we should do the slice after we make the `Set`? so we know we actually get `10`, otherwise if the first 5 are duplicates and we "uniquify" them `getItemIds` will only return 6 no?

RasmusHedvig: Actually, `this.props.items` will never have duplicates by design (so maybe it's not even necessary to do a `Set` then?). And isn't it more efficient to slice 10 items out of total and then map these, instead of mapping all items found (sometimes 1000s) just to ultimately only grab the first 10?

src/features/pricing/index.tsx
RasmusHedvig: At first glance, I see what you mean. But is it _that_ fragile though? Looking at the documentation,

> The Date.parse() method parses a string representation of a date, and returns the number of milliseconds since January 1, 1970, 00:00:00 UTC or NaN if the string is unrecognized or, in some cases, contains illegal date values (e.g. 2015-02-31).

 we have that it actually only returns `NaN` or a `Date` object. `attemptedDate` is always a String, so the worst case we're looking at is actually `NaN` - which we catch on the next line.

src/components/pricing/ItemTable.tsx
palmenhq: Ah, then i think `new Set(..)` could be a bit misleading :)
And in case we don't need to remove duplicates then i think definitely we should slice before. However we're not really doing much heavy lifting as we're just accessing the `id`, so i think performance won't really be an issue even if we have thousands of items to map over :)

src/features/pricing/index.tsx
palmenhq: Yeah i guess its ok :) normally i like to handle these tasks by libraries, so one does not have to think of weird corner cases, but i guess this is quite limited in scope so i think its ok 👍

src/components/pricing/FilterContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+                {itemRow}
+                <Label.Detail>({categoryRow.name})</Label.Detail>
+              </Label>
+            ))}
+
+            {categoryRow.others.length !== 0 ? (
+              <Dropdown text="">
+                <Dropdown.Menu>
+                  {categoryRow.others.map((itemRow) => (
+                    <Dropdown.Item
```
palmenhq: key :)

src/components/pricing/CategoryDropdown.tsx
```diff
@@ -0,0 +1,38 @@
+import { GET_CATEGORIES } from 'features/pricing/queries'
+import * as React from 'react'
+import { Query } from 'react-apollo'
+import { Dropdown } from 'semantic-ui-react'
+
+export default class CategoryDropdown extends React.Component {
```
palmenhq: didn't notice this before but could we avoid default exports and use named exports instead? so this becomes just `export class CategoryDropdown extends ..`

src/components/pricing/ItemTable.tsx
```diff
@@ -0,0 +1,104 @@
...
+        </Table.Header>
+
+        <Table.Body>
+          <Query
+            query={GET_PRICES}
+            variables={{
+              date: this.props.date,
+              ids: this.getItemIds(this.props.items.products),
+            }}
+            fetchPolicy="no-cache"
```
palmenhq: hehe, clever!

PR #68, opened 2019-07-19
Taskmanager (Handle tickets in Hope) [WIP]

fwald:
The first implementation of the Taskmanager. The main tool for tracking and creating tickets. 
Implementation includes fetching tickets from the ticket-service via backoffice and presenting these tickets. The list of tickets can then be sorted and filtered through various options:
- Priority
- Type of ticket
- Status of ticket
- Who the ticket is assigned to

The user can create new tickets and edit existing ones. There is also a rudimentary notification system, using the browser notification API. 

Cheers


src/components/taskmanager-toolbar/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <ToolbarCss>
+        <Grid stackable columns={this.props.items.length}>
+          <Grid.Row style={{ padding: '0.5em' }}>{toolbarItems}</Grid.Row>
+        </Grid>
+      </ToolbarCss>
+    )
+  }
+}
+
+export default Toolbar
```
palmenhq: normally we
a. Don't use default exports
b. Export in the same place as the declaration

So this should probably just be `class Toolbar extends...` without this line :)

palmenhq: Applies to all files

src/components/tickets/ticket/create-ticket/create-ticket.tsx
```diff
@@ -0,0 +1,230 @@
...
+                    <br />
+                    <TextArea
+                      row={4}
+                      col={20}
+                      name="remindMessage"
+                      placeholder="Give a short remind message (max 100 characters)"
+                      value={this.state.remindMessage}
+                      onChange={(e) => this.handleChange(e)}
+                      maxLength={100}
+                      // error={(this.state.remindMessage.length > 0) ? false: true }
```
palmenhq: 🙈

src/components/tickets/ticket/edit-ticket-mut/assignTo.tsx
```diff
@@ -0,0 +1,76 @@
...
+interface IAssignTicketTo {
+  id: string
+  showNotification: (message: string) => void
+  handleChange: (id: string, value: any) => void
+  options: any[]
+  assignedTo: string
+  currentlyAssignedTo: string
+}
+
+export default class AssignTicketToMutation extends React.Component<
```
palmenhq: this could be a [FunctionComponent](https://reactjs.org/docs/components-and-props.html) instead

palmenhq: and i guess that applies to a few places :) we tend to prefer function components here actually

src/components/tickets/ticket/ticketBody.tsx
```diff
@@ -0,0 +1,193 @@
...
+import React from 'react'
+import styled from 'react-emotion'
+import { Button, Divider, Icon, Segment } from 'semantic-ui-react'
+import {
+  createOptionsArray,
+  IEX_TEAM_MEMBERS,
+  lookupStatus,
+  TICKET_STATUS,
+} from '../../../features/taskmanager/types'
+import Notification from '../../notifications/Notification'
```
palmenhq: we do have notifications built already for hedgehog, try to utilise them :)

src/components/tickets/tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+                </p>
+              )
+            }
+
+            // Set up Notifications
+            const [
+              overdueNotifications,
+              upcomingNotifications,
+            ] = this.processReminders(data.tickets)
+            // @TODO Handle notifications that are overdue (unresolved tickets)? Do IEX want to see these?
```
palmenhq: clever 🙌 good comment imo

src/components/tickets/tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+          silent: true,
+          body: reminders[i].remindMessage,
+          icon: null,
+          requireInteraction: true,
+        })
+      }, msUntilFire)
+    }
+  }
+
+  private sortByType = (a, b): number => {
```
palmenhq: you could also do do `(a, b) => b.type - a.type` :)

src/features/taskmanager/types.ts
```diff
@@ -0,0 +1,84 @@
...
+  { text: 'Tomas', value: 'tomas@hedvig.com' },
+  { text: 'Kalle', value: 'karl.jernberg@hedvig.com' },
+  { text: 'Emma', value: 'emma@hedvig.com' },
+  { text: 'Sara', value: 'sara@hedvig.com' },
+  { text: 'Axel', value: 'axel.bryhn@hedvig.com' },
+  { text: 'Unassigned', value: 'Unassigned' },
+]
+
+// Purely for making it easier to read in the UI:
+const IEX_TEAM_NAME_LOOKUP = {
```
palmenhq: Maybe you could do something like
```js
const IEX_TEAM_NAME_LOOKUP = Object.keys(IEX_TEAM_MEMBERS).reduce((acc, key) => ({ ...acc, [IEX_TEAM_MEMBERS[key]]: key }))
```

So we don't have to maintain them in 2 places

src/store/types/ticket.ts
```diff
@@ -0,0 +1,15 @@
+export interface ITicket {
+    ticketId: string,
+    issuedBy: string,
+    assignedTo: string,
+   type: string,
+    remind: RemindInfo, 
+    priority: number,
+    description: string,
+}
+//TODO: Add tags
```
palmenhq: 👀

src/components/taskmanager-toolbar/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  boxShadow: '1px 1px 5px lightgray',
+  marginBottom: '3em',
+  padding: '0',
+})
+
+
+class Toolbar extends React.Component<IToolbar, {}> {
+  public render() {
+    const toolbarItems = this.props.items.map((item) => (
+      <Grid.Column key={item.id + 'g'}>
```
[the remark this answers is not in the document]
fwald: Mainly because the ToolbarItem also gets key= ticket.id and I thought that keys had to be unique?

src/components/tickets/ticket/create-ticket/datepicker.tsx
```diff
@@ -0,0 +1,49 @@
...
+    value: any
+  }
+  timepicker: {
+    name: string
+    value: any
+  }
+  handleChange: (event: any) => void
+}
+
+class Datepicker extends React.Component<IDatepicker, {}> {
```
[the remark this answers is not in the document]
fwald: Yes I looked at that. If i remember correctly it only selects for 'Date' but I also needed time so I thought i would bundle them together. I can use the existing Datepicker for the date portion though?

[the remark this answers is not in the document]
fwald: I should have called the component DateTimePicker though =)

src/components/tickets/tickets.tsx
fwald: Does this work for comparison of strings?

src/components/tickets/ticket/edit-ticket-mut/assignTo.tsx
fwald: I've changed the Mutation Components (assignTo, description, reminder, status). They now use Redux for notifications. Therefore it seems easier to keep them as class based components? I will change other components though to make them FunctionalComponents.

src/components/taskmanager-toolbar/index.tsx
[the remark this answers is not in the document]
palmenhq: Ah, i think no because React uses the key to keep track of array items that are "neighbors" so when it compares the tree once you change things it'll understand that the two components are different. Also coming back to this i think you don't need the `key` prop on `ToolbarItem` at all since it's only used to compare array items to each other, and it's not part of an array, it's just a normal child :)

src/components/tickets/ticket/create-ticket/create-ticket.tsx
```diff
@@ -0,0 +1,211 @@
...
+                        remindNotificationDate: this.state.remindDate,
+                        remindNotificationTime: this.state.remindTime,
+                        remindMessage: this.state.remindMessage,
+                        description: this.state.description,
+                        status: 'WAITING',
+                      },
+                    },
+                    refetchQueries: [{ query: GET_TICKETS }],
+                  })
+                  this.props.closeModal()
```
palmenhq: i think actually we should do this once `createNewTicket` finishes. It's asynchronous so you can't really trust that it's done at this stage. Rather it returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) and imo you should close the modal once the mutation finishes.

src/components/tickets/ticket/create-ticket/create-ticket.tsx
```diff
@@ -0,0 +1,211 @@
...
+                    <Divider />
+                    <Label htmlFor={'remindMessage'}>Message:</Label>
+                    <br />
+                    <TextArea
+                      row={4}
+                      col={20}
+                      name="remindMessage"
+                      placeholder="Give a short remind message (max 100 characters)"
+                      value={this.state.remindMessage}
+                      onChange={(e) => this.handleChange(e)}
```
palmenhq: i think in this case `onChange={this.handleChange}` should be sufficient, no need to make an arrow function

src/components/tickets/ticket/edit-ticket-mut/assignTo.tsx
```diff
@@ -0,0 +1,94 @@
...
+  handleChange: (id: string, value: any) => void
+  options: any[]
+  assignedTo: string
+  currentlyAssignedTo: string
+}
+
+class AssignTicketToMutation extends React.Component<IAssignTicketTo, {}> {
+  public render() {
+    return (
+      <Mutation mutation={ASSIGN_TO} key={this.props.id + 'assign'}>
```
palmenhq: no need for the `key` prop here i believe

src/components/tickets/ticket/edit-ticket-mut/assignTo.tsx
```diff
@@ -0,0 +1,94 @@
...
+            <Form
+              onSubmit={(e) => {
+                e.preventDefault()
+                assignTicketToTeamMember({
+                  variables: {
+                    ticketId: this.props.id,
+                    teamMemberId: this.props.assignedTo,
+                  },
+                })
+                  .then(() => {
```
palmenhq: nice!

src/components/tickets/ticket/edit-ticket-mut/assignTo.tsx
```diff
@@ -0,0 +1,94 @@
...
+                      type: 'green',
+                    })
+                  })
+                  .catch((error) => {
+                    this.props.showNotification({
+                      header: 'Error',
+                      message: error.message,
+                      type: 'red',
+                    })
+                    throw error
```
palmenhq: very nice, then sentry can catch this if we turn it on for hegehog some time

src/components/tickets/ticket/edit-ticket-mut/description.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+class ChangeDescriptionMutation extends React.Component<
+  IChangeDescription,
+  {}
+> {
+  public render() {
+    return (
+      <Mutation
+        mutation={CHANGE_DESCRIPTION}
+        key={this.props.id + 'description'}
```
palmenhq: same, no need for `key` :)

src/components/tickets/ticket/edit-ticket-mut/reminder.tsx
```diff
@@ -0,0 +1,118 @@
...
+  remindMessage: string
+  showNotification: (data: any) => void
+  handleChange: (event: any) => void
+  currentReminder: IRemindNotification
+}
+
+class ChangeReminderMutation extends React.Component<IChangeReminder, {}> {
+  public render() {
+    return (
+      <Mutation mutation={CHANGE_REMINDER} key={this.props.id + 'reminder'}>
```
palmenhq: key :D

src/components/tickets/ticket/edit-ticket-mut/reminder.tsx
```diff
@@ -0,0 +1,118 @@
...
+              <p>Current reminder: {this.reminderToString()}</p>
+            </Form>
+          )
+        }}
+      </Mutation>
+    )
+  }
+  private reminderToString = () => {
+    if (
+      this.props.currentReminder === undefined ||
```
palmenhq: `!this.props.currentReminder || !this.props.currentReminder.date ...` is ok also, so you don't need to perform so many checks :)

src/components/tickets/ticket/ticket.tsx
```diff
@@ -0,0 +1,146 @@
...
+          columns="equal"
+          verticalAlign="middle"
+        >
+          <Grid.Row>
+            <Grid.Column floated="left">
+              <Popup
+                content={'Ticket type: ' + this.props.type.toLowerCase()}
+                trigger={<span>{this.getTypeIcon(this.props.type)}</span>}
+              />
+              {this.props.status === TicketStatus.RESOLVED ? (
```
palmenhq: just for future reference `{this.props.status === TicketStatus.RESOLVED && <Popup ... />}` is ok also and you can save yourself a few keystrokes

src/components/tickets/ticket/ticket.tsx
```diff
@@ -0,0 +1,146 @@
...
+    this.setState({ showBody: updatedState })
+  }
+
+  private getTypeIcon = (type) => {
+    const icon = typeIcons[type]
+    return icon
+  }
+}
+
+const getReminderTimeInWords = (date, time) => {
```
palmenhq: i'm not sure how detailed this needs to be but i guess we could also use something like [distanceInWordsToNow](https://date-fns.org/v1.30.1/docs/distanceInWordsToNow)

palmenhq: (not saying you should but it _could_ be useful)

src/components/tickets/tickets.tsx
palmenhq: No, i thought the type was an ordinal enum or something. Does this work for comparison of strings though? 😬 i feel like the order shouldn't necessarily be alphabetical?

src/features/taskmanager/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+    const sortCategory = id
+    const sortOrder = this.changeOrder(
+      id,
+      this.state.sort.category,
+      this.state.sort.order,
+    )
+
+    const toolbarItemsToUpdate = [...this.state.toolbarItems]
+
+    for (let i = 0; i < toolbarItemsToUpdate.length; i++) {
```
palmenhq: could we rewrite this with [`Array#map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) instead? I don't like for loops and mutability 😬

src/features/taskmanager/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+      }
+      if (id !== this.state.sort.category) {
+        if (toolbarItemsToUpdate[i].id === this.state.sort.category) {
+          const deactivateItem = { ...toolbarItemsToUpdate[i], isActive: false } // need to deactivate the previously active item
+          toolbarItemsToUpdate[i] = deactivateItem
+        }
+      }
+    }
+
+    const sort = { ...this.state.sort }
```
palmenhq: you could also just do
```js
const sort = {
  ...this.state.sort,
  order: sortOrder,
  category: sortCategory,
}
```

src/features/taskmanager/types.ts
```diff
@@ -0,0 +1,79 @@
...
+  { text: 'Kalle', value: 'karl.jernberg@hedvig.com' },
+  { text: 'Emma', value: 'emma@hedvig.com' },
+  { text: 'Sara', value: 'sara@hedvig.com' },
+  { text: 'Axel', value: 'axel.bryhn@hedvig.com' },
+  { text: 'Unassigned', value: 'Unassigned' },
+]
+
+// Purely for making it easier to read in the UI:
+let IEX_TEAM_NAME_LOOKUP = {}
+IEX_TEAM_MEMBERS_OPTIONS.forEach(
```
palmenhq: plz use [`Array#reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) instead

src/features/taskmanager/types.ts
```diff
@@ -0,0 +1,79 @@
...
+  (entry) =>
+    (IEX_TEAM_NAME_LOOKUP = {
+      ...IEX_TEAM_NAME_LOOKUP,
+      [entry.value]: entry.text,
+    }),
+)
+
+export const lookupTeamMemberName = (email: string): string => {
+  name = IEX_TEAM_NAME_LOOKUP[email]
+  if (name === '' || name === 'undefined') {
```
palmenhq: `typeof name === 'undefined'` no?
I guess we could also just do `return IEX_TEAM_NAME_LOOKUP[email] || 'Unassigned'`', it's fine by me

src/components/taskmanager-toolbar/index.tsx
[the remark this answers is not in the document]
fwald: True =D

src/components/tickets/tickets.tsx
fwald: Well I changed it to filter by type instead. Think that makes more sense. This means that this function is no longer necessary

src/features/taskmanager/index.tsx
fwald: OFC

src/components/tickets/ticket/ticket.tsx
fwald: That is actually the one I wanted to use. But I could not find it in the module... After digging around I discovered that they changed the name to formatDistance() in later versions.. Will use that one

src/features/taskmanager/types.ts
```diff
@@ -0,0 +1,80 @@
...
+  { text: 'Emma', value: 'emma@hedvig.com' },
+  { text: 'Sara', value: 'sara@hedvig.com' },
+  { text: 'Axel', value: 'axel.bryhn@hedvig.com' },
+  { text: 'Unassigned', value: 'Unassigned' },
+]
+
+// Purely for making it easier to read in the UI:
+const IEX_TEAM_NAME_LOOKUP = IEX_TEAM_MEMBERS_OPTIONS.reduce(
+  (acc, val) => {
+    return { ...acc, [val.value]: val.text }
```
palmenhq: beautiful 😍

PR #70, opened 2019-07-23
discrete colors and hide autocomplete when autoanswer

hkindbom:
ready to merge to dev, also merge autoanswer service

### DIFF
--- a/src/components/chat/tabs/ChatTab.tsx
+++ b/src/components/chat/tabs/ChatTab.tsx
@@ -88,7 +88,7 @@ export default class ChatTab extends React.Component {
 
     const lastMemberMessagesArray = messages.filter(
       (message, index) =>
-        ['free.chat.message', 'free.chat.from.bo', 'free.chat.start'].includes(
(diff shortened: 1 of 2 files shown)


src/components/chat/tabs/ChatTab.tsx
```diff
@@ -88,7 +88,7 @@ export default class ChatTab extends React.Component {
 
     const lastMemberMessagesArray = messages.filter(
       (message, index) =>
-        ['free.chat.message', 'free.chat.from.bo', 'free.chat.start'].includes(
+        ['free.chat.message', 'free.chat.from.bo'].includes(
```
palmenhq: is this intentional?

hkindbom: yea, i think it it unneccessary but we can keep it if you would like? It will not cause any issues

palmenhq: nah if its not used lets drop it

PR #78, opened 2019-08-09
Ticket history page

fwald:
A new feature for Hedgehog where we can see tall the revisions that have been made to a ticket.
Also did some minor changes to the UI of taskmanager.

### DIFF
--- a/src/components/tickets/ticket/ticketBody.tsx
+++ b/src/components/tickets/ticket/ticketBody.tsx
@@ -9,6 +9,7 @@ import {
   Icon,
   Segment,
 } from 'semantic-ui-react'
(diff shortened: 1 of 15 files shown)


src/components/tickets/ticket/ticketBody.tsx
```diff
@@ -215,6 +215,8 @@ export class TicketBody extends React.Component<ITicketBody, ITicketBodyState> {
       </>
     )
 
+    const ticketId = this.props.id
```
AlexSJ1990: when do we use this?

fwald: oopsie, removed it

PR #80, opened 2019-08-20
Fix modify insurance date bug

palmenhq:
Previously we instantly reloaded the page after the button was clicked = if the endpoint wasn't fast enough the request wouldn't finish before it was aborted due to page reload.


src/store/sagas/insuranceSaga.ts
```diff
@@ -163,6 +163,7 @@ function* modifyInsuranceFlow({ memberId, request }) {
     )
 
     yield put(modifyInsuranceSuccess(response.data))
+    window.location.reload() // Sorry
```
vonElfvin: Why Sorry? 😅

PR #82, opened 2019-08-21
Feature/syna tab

AlexSJ1990:


src/components/chat/tabs/DebtTab.tsx
```diff
@@ -0,0 +1,341 @@
...
+        <Table.HeaderCell>Claimant</Table.HeaderCell>
+      </Table.Row>
+    </Table.Header>
+    <Table.Body>
+      {[...paymentDefaults]
+        .sort(sortPaymentDefaultByYear)
+        .map((paymentDefault) => (
+          <Table.Row
+            key={
+              paymentDefault.year.toString() +
```
palmenhq: i don't know exactly what `caseId` is but it sounds like what we would want to use here no?

src/components/chat/tabs/DebtTab.tsx
```diff
@@ -0,0 +1,341 @@
+import { MonetaryAmount } from '../../../components/claims/claim-details/components/ClaimPayments'
```
palmenhq: we should move this component to a more generic place

src/components/chat/tabs/DebtTab.tsx
```diff
@@ -0,0 +1,341 @@
...
+        <Table.HeaderCell>Number of Public Debts</Table.HeaderCell>
+        <Table.HeaderCell>Total Amount of Private Debt</Table.HeaderCell>
+        <Table.HeaderCell>Number of Private Debts</Table.HeaderCell>
+        <Table.HeaderCell>Date of Debt Check</Table.HeaderCell>
+      </Table.Row>
+    </Table.Header>
+    <Table.Body>
+      <Table.Row>
+        <Table.Cell>
+          {debt.totalAmountPublicDebt.amount}{' '}
```
palmenhq: imo we should use the money formatter instead. 
https://github.com/HedvigInsurance/hedgehog/blob/5058d707d94594186e915f851a895b9b88a59683/src/lib/intl.ts#L6-L12

src/components/chat/tabs/DebtTab.tsx
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  private confirm = () => {
+    this.setState({ confirming: true })
+  }
+}
+
+const mapActions = { ...actions.notificationsActions }
+
+export default MemberDebtComponent
```
palmenhq: no default exports please :)

src/components/chat/tabs/DebtTab.tsx
```diff
@@ -0,0 +1,341 @@
...
+                        },
+                      ]}
+                    >
+                      {(mutation, { loading }) => {
+                        return (
+                          <Button
+                            disabled={loading}
+                            onClick={
+                              this.state.confirming
+                                ? () => {
```
palmenhq: Maybe we could move this to a separate function? Inlining it like this works but gets a bit hard to read imo. Something like what we do with `resetButton`

src/components/chat/tabs/DebtTab.tsx
AlexSJ1990: showed it to the iex team, and they said they didn't need it - so I shall remove from the query 👍

src/components/chat/tabs/DebtTab.tsx
AlexSJ1990: yes, this looks so much nicer =)

src/components/chat/tabs/DebtTab.tsx
```diff
@@ -0,0 +1,328 @@
...
+    <Table.Header>
+      <Table.Row>
+        <Table.HeaderCell>Year</Table.HeaderCell>
+        <Table.HeaderCell>Payment Default Type</Table.HeaderCell>
+        <Table.HeaderCell>Debt Amount</Table.HeaderCell>
+        <Table.HeaderCell>Claimant</Table.HeaderCell>
+      </Table.Row>
+    </Table.Header>
+    <Table.Body>
+      {[...paymentDefaults]
```
palmenhq: muy bien

src/components/chat/tabs/DebtTab.tsx
```diff
@@ -0,0 +1,328 @@
...
+  background: colors.GREEN,
+  color: colors.WHITE,
+  fontWeight: 600,
+  border: 0,
+  borderRadius: 5,
+  fontSize: '1.2rem',
+  transition: 'background 250ms',
+
+  '&:hover, &:focus': {
+    background: '#18cea1',
```
palmenhq: can we get this from @hedviginsurance/brand?

PR #97, opened 2019-10-07
Visualize charge status under Account tab

akashpatl:
Yellow for deducted amount and red for failed charge. Also added column for type of charge.

### DIFF
--- a/src/components/chat/tabs/AccountTab.tsx
+++ b/src/components/chat/tabs/AccountTab.tsx
@@ -35,6 +35,9 @@ export const GET_MEMBER_ACCOUNT_QUERY = gql`
           title
           source
           reference
+          type
+          failedAt
+          chargedAt
         }
       }
     }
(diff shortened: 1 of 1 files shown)


src/components/chat/tabs/AccountTab.tsx
```diff
@@ -45,9 +48,17 @@ export interface AccountTabProps {
...
+const TableRowColored = styled(TableRow)(({ entry }: { entry }) => {
+  let backgroundColor
+  if (entry.failedAt) {
+    backgroundColor = '#FFDDDD'
+  } else if (entry.amount.amount < 0) {
+    backgroundColor = '#FFFFDD'
+  } else {
+    backgroundColor = '#DDFFDD'
+  }
+  return { backgroundColor }
```
palmenhq: just to avoid mutability and `else` clauses we could just do `return { backgroundColor: '<color>' }` in each `if` :)
imo that's slightly neater

akashpatl: Great! Thanks!

PR #99, opened 2019-10-21
Feature/more info claim

vonElfvin:


src/utils/DateUtil.ts
```diff
@@ -0,0 +1,61 @@
+export function timeAgo(pastDate: Date) {
+  const date = new Date(pastDate)
+  const today = new Date()
+  if (today.getMilliseconds() - date.getMilliseconds() > 0) {
+    throw Error('Cannot get timeAgo for date in the future')
+  }
+  return dateDifference(date, today)
+}
+
+function dateDifference(startingDate, endingDate) {
```
palmenhq: I _really_ don't think we should maintain code like this ourselves. There's a [date-fns function](https://date-fns.org/docs/distanceInWords) that does exactly this, but slightly better, date-fns is aleady a dependency of hope..

vonElfvin: Cool, didn't find any example when looking. Changed it now 👍

src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -77,15 +96,34 @@ const MemberInformation: React.SFC<MemberInformationProps> = ({
...
+      <b>Personal Number:</b> {personalNumber}
+    </p>
+    <p>
+      <b>Address:</b> {address}, {postalNumber} {city}
+    </p>
+    <p>
+      <b>Failed Charge Spree:</b> {numberFailedCharges}
+    </p>
+    <p>
+      <b>Total Debt (Minimum):</b> {Math.trunc(amount)} {currency}
```
palmenhq: Could we use [`formatMoney`](https://github.com/HedvigInsurance/hedgehog/blob/master/src/lib/intl.ts#L12) for this?

src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -1,3 +1,5 @@
+import {formatDistance} from "date-fns"
```
palmenhq: prettier :)

PR #101, opened 2019-10-23
Feature/backfill accountab manually

akashpatl:


src/components/chat/tabs/account-tab/add-entry-form.tsx
```diff
@@ -27,10 +27,11 @@ interface State {
   confirmed: boolean
 }
 
-const SubmitButton = withStyles({
+export const SubmitButton = withStyles({
```
palmenhq: If we need this in more places, maybe we should move the whole thing so some “common” module?

PR #102, opened 2019-11-07
Eject from web-survival-kit and upgrade stuff

palmenhq:
- Ejects web-survival-kit to ease understanding what webpack does here
- Add support for [TypeScript 3.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html)
- Adds support for react 16.11 (including hooks and suspense support)
- Improves initial development build time from 46 seconds to 25 seconds
- The server no longer restarts on every change
- Decreases bundle size by 20% by lazy-loading emoji mart

Still waiting for the prettier 1.19 release (that supports TS 3.7 but i believe we can merge this in the meantime and upgrade later)

The life of hedgehog developers just got a whole lot better


src/components/chat/chat/EmojiPicker.tsx
```diff
@@ -29,12 +29,14 @@ const actions: ActionMap<State, Actions> = {
   setIsOpen: (open: boolean) => () => ({ open }),
 }
 
-const LazyEmojiPicker = React.lazy(() =>
-  import(/* webpackChunkName: 'emoji-mart' */ 'emoji-mart').then(
-    ({ Picker }) => Picker,
-  ) as any,
+const LazyEmojiPicker = React.lazy(
```
sampettersson: whut, didn't know `React.lazy` was a thing now, cool stuff!

palmenhq: Its amaze!!

PR #105, opened 2019-11-18
Feature/itemizer

akashpatl:


src/components/claims/claim-details/components/inventory/InventoryEntry.tsx
```diff
@@ -0,0 +1,181 @@
...
+            {({ loading, data, error }) => {
+              if (loading) {
+                return null
+              }
+              if (error) {
+                return null
+              }
+
+              if (data) {
+                if (!data.inventoryItemFilters) {
```
palmenhq: We could now utilise TS 3.7  and optional chaining like so:
```ts
if (!data?.inventoryItemFilters) {
  return null
}
```

src/components/pricing/ItemTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+  public handlePriceData = (data) => {
+    if (data && 'prices' in data) {
+      return data.prices.reduce(
+        (accumulator, s) => ({ ...accumulator, [s.itemId]: s }),
+        {},
+      )
+    }
+
+    return {}
+  }
```
palmenhq: This could now become a oneliner like so

```ts
data => data?.prices?.reduce(...) ?? {}
```

all thanks to ts 3.7 :D

src/features/pricing/index.tsx
```diff
@@ -0,0 +1,280 @@
...
+import { Query } from 'react-apollo'
+import { Dropdown, Header, Icon, Input } from 'semantic-ui-react'
+
+export default class Pricing extends React.Component {
+  public state = {
+    activeCategory: '',
+    activeCategoryName: '',
+    activeQuery: '',
+    activeFilters: [],
+    activeDate: new Date().toJSON().slice(0, 10),
```
palmenhq: please use a proper date formatter, `import { format } from 'date-fns'`  is a really good condidate

src/components/claims/claim-details/components/inventory/ClaimItemDatabase.tsx
```diff
@@ -43,6 +43,7 @@ export class ClaimItemDatabase extends React.Component {
         <Grid item xs={12} sm={12} md={4}>
           <Query
             query={GET_INVENTORY}
+            pollInterval={2500}
```
palmenhq: Is this intentional? Whats the reasoning for polling?

src/components/claims/claim-details/components/inventory/EditableCategoryDropdown.tsx
```diff
@@ -0,0 +1,44 @@
...
+              options={
+                loading || error
+                  ? [{ key: 1, text: 'None', value: 1 }]
+                  : data.categories.map((item) => {
+                      return {
+                        key: item.id,
+                        text: item.name.split(' > ')[
+                          item.name.split(' > ').length - 1
+                        ],
+                        value: item.id,
```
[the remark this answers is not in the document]
akashpatl: This is a really confusing data parser where the item service is passing the actual category string from the original data source. This should most likely be cleaned up in a new iteration where we change how item categories are mapped. However, breaking out to a separate function should be done now.

src/components/claims/claim-details/components/inventory/AddItemPrompt.tsx
```diff
@@ -0,0 +1,238 @@
...
+        >
+          Cancel
+        </Button>
+
+        <Query
+          query={GET_ALL_FILTERS}
+          variables={{
+            categoryId: this.state.itemCategory
+              ? this.state.itemCategory
+              : '-1',
```
palmenhq: is this intentional? A string number?

src/components/claims/claim-details/components/inventory/AddItemPrompt.tsx
```diff
@@ -0,0 +1,238 @@
...
+                </Button>
+
+                  { this.state.showFilters && loading &&
+                  <Loader
+                    active
+                    inline="centered"
+                    style={{ marginTop: '20px' }}
+                  />
+                }
+                  { this.state.showFilters && !loading &&
```
palmenhq: can we run prettier on this file please :)

src/components/claims/claim-details/components/inventory/ClaimItemDatabase.tsx
palmenhq: I really think we should remove this now that we added a refetch query instead

src/components/pricing/CategoryDropdown.tsx
```diff
@@ -0,0 +1,41 @@
+import { GET_CATEGORIES } from 'features/pricing/queries'
+import * as React from 'react'
+import { Query } from 'react-apollo'
+import { Dropdown } from 'semantic-ui-react'
+
+export class CategoryDropdown extends React.Component {
+  //TODO: Pass clean category names from item-service instead of parsing here
+  public parseCategoryName(text) {
```
palmenhq: for future reference i think we should make these type of functions pure standalone funcs, but this is ok for now

akashpatl: I agree! Although I think this function should be removed in the future and the item service should take care of sending cleaner data

PR #106, opened 2019-11-21
Feature/add upload claim file button

AlexSJ1990:


src/components/claims/claim-details/components/ClaimFileTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+                                {claimFile.category != null
+                                  ? claimFile.category
+                                  : 'File Type'}
+                              </option>
+                              <option>Reciept</option>
+                              <option>Proof of ownership</option>
+                              <option>Police report</option>
+                              <option>Invoice</option>
+                              <option>Proof of damage</option>
+                              <option>Other</option>
```
palmenhq: maybe we should add some proper `value` properties?

src/components/claims/claim-details/components/FileUpload.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+const UploadClaimFileHeader = styled('h3')({})
+
+export const Button = styled('button')({
+  display: 'inline-flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  fontSize: 'inherit',
+  font: 'inherit',
+  padding: '4rem 4rem',
```
palmenhq: not that much of a biggie but i guess `padding: '4rem'` would be sufficient

src/components/claims/claim-details/components/FileUpload.tsx
```diff
@@ -0,0 +1,124 @@
...
+  border: '1px solid ',
```
palmenhq: no `borderColor` this won't give any effect unless we actually set either `border: '1px solid <the color>'` or a `borderColor` property?

src/components/claims/claim-details/components/FileUpload.tsx
```diff
@@ -0,0 +1,124 @@
...
+    textDecoration: 'none',
+  },
+  '&:focus, &:hover': {
+    outline: 'none',
+    boxShadow: 'none',
+  },
+})
+
+const FileUploadContainer = styled('div')({
+  padding: '2rem 2rem',
```
palmenhq: same here, `2rem` should do it :)

src/components/claims/claim-details/components/FileUpload.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  render() {
+    return (
+      <>
+        <UploadClaimFileWrapper>
+          <UploadClaimFileHeader>Upload files here</UploadClaimFileHeader>
+          <FileUploadContainer>
+            <Dropzone
+              onDrop={(files) => {
```
palmenhq: we could just do `onDrop={this.onDrop}` also :)

src/components/claims/claim-details/components/FileUpload.tsx
```diff
@@ -0,0 +1,124 @@
...
+class FileUploadComponent extends React.Component<{
+  claimId: string
+  memberId: string
+  showNotification: (data: any) => void
+  onUploaded: () => void
+}> {
+  public state = {
+    value: null,
+    uploaded: false,
+  }
```
palmenhq: do we need this? We're not really reading it anywhere are we?

src/components/claims/claim-details/index.tsx
```diff
@@ -78,68 +94,111 @@ interface Props {
...
+                reserves={reserves}
+                sanctionStatus={member.sanctionStatus}
+              />
+            </Grid>
+            <Grid item xs={12}>
+              <Query
+                query={CLAIM_FILES_QUERY}
+                variables={{ id: match.params.id }}
+              >
+                {({ loading, error, data, refetch }) => {
```
palmenhq: these'll be shadowed now no? I think we should not destruct the object if we nest the queries.
An option would also be to merge the queries into one - it's a nice feature of graphql to be able to do that

src/components/claims/claim-details/components/DeleteClaimFileButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+  private handleClick = (mutation) => {
+    mutation({
+      variables: {
+        claimId: this.props.claimId,
+        claimFileId: this.props.claimFileId,
+      },
+    })
+      .then(() => {
+        this.props.showNotification({
+          message: 'file has been marked as deleted',
```
vonElfvin: From a UX perspective the file has been deleted I guess:
```suggestion
          message: 'File has been deleted',
```

src/components/claims/claim-details/index.tsx
AlexSJ1990: good point, we can get the claimFiles from the claim query so no need for this query at all!

src/components/claims/claim-details/components/ClaimFileTable.tsx
```diff
@@ -0,0 +1,153 @@
...
+                          />
+                        ) : (
+                          <Image src={claimFile.fileUploadUrl} size="large" />
+                        )}
+                      </Table.Cell>
+                      <Table.Cell>
+                        <Mutation mutation={SET_CLAIM_FILE_CATEGORY}>
+                          {(mutation) => {
+                            return (
+                              <select
```
palmenhq: Maybe we could use `semantic-ui`'s dropdown?

src/components/claims/claim-details/components/ClaimFileTable.tsx
```diff
@@ -0,0 +1,153 @@
...
+                                class="ui fluid search dropdown"
+                                multiple=""
+                                onChange={() => this.handleSelect(event, mutation, claimFile.claimFileId)}
```
palmenhq: imo we could do this inline instead of abstracting it to a function, in this particular case i think it just makes the code slightly harder to folow

src/components/claims/claim-details/components/ClaimFileTable.tsx
```diff
@@ -0,0 +1,153 @@
...
+                            src={claimFile.fileUploadUrl}
+                            width="800px"
+                            height="300px"
+                          />
+                        ) : (
+                          <Image src={claimFile.fileUploadUrl} size="large" />
+                        )}
+                      </Table.Cell>
+                      <Table.Cell>
+                        <Mutation mutation={SET_CLAIM_FILE_CATEGORY}>
```
palmenhq: just an idea, should we use the hooks api instead?

src/components/claims/claim-details/components/ClaimFileTable.tsx
```diff
@@ -0,0 +1,153 @@
...
+            </Table.Row>
+          </Table.Header>
+          <Table.Body>
+            {this.props.claimFiles.length === 0 ? (
+              <NoClaimFiles>
+                No claim documents have been uploaded for this claim
+              </NoClaimFiles>
+            ) : (
+              [...this.props.claimFiles]
+                .sort(sortClaimFileDate)
```
palmenhq: amaze!

src/components/claims/claim-details/components/DeleteClaimFileButton.tsx
```diff
@@ -0,0 +1,108 @@
...
+                Delete
+              </Button>
+            ) : (
+              <Button
+                disabled={loading}
+                onClick={() => this.handleClick(mutation)}
+              >
+                Are you sure?
+              </Button>
+            )}
```
palmenhq: imo this UX can be a bit confusing, maybe we should just do good-ol-fashioned `confirm` instead?

src/components/claims/claim-details/components/DeleteClaimFileButton.tsx
```diff
@@ -0,0 +1,108 @@
...
+    })
+      .then(() => {
+        this.props.showNotification({
+          message: 'File has been deleted',
+          header: 'Success',
+          type: 'olive',
+        })
+        this.resetConfirmed()
+      })
+      .catch((error) => {
```
palmenhq: maybe we should rethrow this? Just to be sure this error is visible in the console or sentry?

src/components/claims/claim-details/components/FileUpload.tsx
```diff
@@ -0,0 +1,112 @@
...
+      body: claimFiles,
+    })
+      .then(() => {
+        this.props.showNotification({
+          message: 'upload successful!',
+          header: 'Approved',
+          type: 'olive',
+        })
+        this.props.onUploaded()
+        window.location.reload()
```
palmenhq: hm. any chance we could just make some refetch queries rather? I think this should really really be our last way out when there's no other way of refetching all state properly

src/components/claims/claim-details/components/ClaimFileTable.tsx
palmenhq: (for all places, in that case, ofc)

src/components/claims/claim-details/components/ClaimFileTable.tsx
```diff
@@ -0,0 +1,171 @@
...
+import * as React from 'react'
+import { Mutation } from 'react-apollo'
+import styled from 'react-emotion'
+
+import { Table, Image } from 'semantic-ui-react'
+import actions from '../../../../store/actions'
+import { connect } from 'react-redux'
+import { DeleteButton } from '../components/DeleteClaimFileButton'
+import { dateTimeFormatter } from '../../../../lib/helpers'
+import { Dropdown } from 'semantic-ui-react'
```
palmenhq: shall we maybe move this to L6? :) feels a bit confusing to import the same module multiple times

PR #107, opened 2019-11-21
Feature/scroll chat to bottom

vonElfvin:
* Scrolls the chat to the buttom when the chat is opened for the first time and every time the member writes a message, after consultation with IEX this is the prefered behaviour.

### DIFF
--- a/src/components/chat/messages/MessagesList.tsx
+++ b/src/components/chat/messages/MessagesList.tsx
@@ -1,17 +1,18 @@
+import animateScrollTo from 'animated-scroll-to'
 import Message from 'components/chat/messages/Message'
 import * as PropTypes from 'prop-types'
 import * as React from 'react'
-import styled from 'styled-components'
+import styled from 'react-emotion'
 
-const MessagesListContainer = styled.div`
-  box-sizing: border-box;
-  overflow-y: auto;
-  padding: 20px 20px 60px;
-`
+const MessagesListContainer = styled('div')({
+  boxSizing: 'border-box',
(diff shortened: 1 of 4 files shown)


src/components/chat/messages/MessagesList.tsx
```diff
@@ -22,18 +23,20 @@ export default class MessagesList extends React.Component {
...
-      const msgNode = document.getElementById(`msg-${id}`)
-      if (id && msgNode) {
-        msgNode.scrollIntoView()
+      const { messages } = this.props
+      const lastMessage = messages[messages.length - 1]
+      const lastMessageElement = document.getElementById(
+        `msg-${lastMessage.globalId}`,
+      )
+      if (lastMessageElement) {
+        animateScrollTo(lastMessageElement, { elementToScroll: list })
```
palmenhq: i think we should
a) add a sanity check so the messages length is actually different just to not cause any issues if we start using state in this component (`componentDidUpdate` is called also on state update)
b) clean this up a bit, probably we don't need the timeout now that we're making the scroll _after_ the update

src/components/chat/messages/MessagesList.tsx
```diff
@@ -1,17 +1,18 @@
+import animateScrollTo from 'animated-scroll-to'
 import Message from 'components/chat/messages/Message'
 import * as PropTypes from 'prop-types'
 import * as React from 'react'
-import styled from 'styled-components'
+import styled from 'react-emotion'
```
palmenhq: 👩‍🎤😍

PR #108, opened 2019-11-21
Fix/fraud status in claims

vonElfvin:
Depends on https://github.com/HedvigInsurance/hedgehog/pull/107


src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -114,7 +114,7 @@ const MemberInformation: React.SFC<MemberInformationProps> = ({
     <p style={{ marginTop: '-7px' }}>
       <b>Fraudulent Status:</b>{' '}
       <span style={{ fontSize: '32px' }}>
-        <FraudulentStatus stateInfo={fraudulentStatus} />
+        <FraudulentStatus stateInfo={{ state: fraudulentStatus }} />
```
vonElfvin: this is the only line that is changed

PR #110, opened 2019-11-22
Resolve issues after date-fns upgrade

palmenhq:


src/components/chat/tabs/account-tab/add-entry-form.tsx
```diff
@@ -122,7 +120,7 @@ export class AddEntryForm extends React.Component<
                       amount: parseAmount(formData.amount),
                       currency: 'SEK',
                     },
-                    fromDate: format(formData.fromDate, 'yyyy-MM-dd'),
+                    fromDate: format(parseISO(formData.fromDate), 'yyyy-MM-dd'),
```
palmenhq: this looks weird, `fromDate` is already a date no?

PR #117, opened 2019-12-19
add tickets to claims view and member page

AlexSJ1990:


src/components/claims/claim-details/index.tsx
```diff
@@ -150,21 +156,17 @@ const ClaimPage: React.SFC<Props> = ({ match }) => (
...
-                />
-              </Grid>
-              <Grid item xs={12}>
-                <ClaimFileTable
-                  claimFiles={claimFiles}
-                  claimId={match.params.id}
-                />
-              </Grid>
-            </>
+            <CreateTicketStandAlone claimId={match.params.id} />
```
vonElfvin: This could probably be named referenceId instead
```suggestion
            <CreateTicketStandAlone referenceId={match.params.id} />
```

PR #121, opened 2020-01-02
Wait for claims service to (hopefully) finish before refetching claims page

palmenhq:


src/components/claims/claim-details/components/ClaimType.tsx
```diff
@@ -398,180 +399,188 @@ const handleError = (showNotification: (data: any) => void) => () => {
...
-                  setClaimInformation({
-                    variables: {
-                      id: claimId,
-                      claimInformation: {
-                        ...values,
-                        date: values.date && format(values.date, 'yyyy-MM-dd'),
-                      },
-                    },
+}> = ({ type, claimId, refetchPage, showNotification }) => {
+  const [isSetClaimTypeLoading, setSetClaimTypeLoading] = React.useState(false)
```
vonElfvin: not just useState?

src/components/claims/claim-details/components/ClaimType.tsx
```diff
@@ -398,180 +399,188 @@ const handleError = (showNotification: (data: any) => void) => () => {
...
+                initialValues={{ selectedType: type?.__typename || '' }}
+                onSubmit={async (values, { setSubmitting }) => {
+                  if (!values.selectedType) {
+                    return
+                  }
+                  setSetClaimTypeLoading(true)
+                  await setClaimType({
+                    variables: { id: claimId, type: values.selectedType },
                   })
+                  await sleep(2000) // wait for claims service to propagate changes :(
```
vonElfvin: :(

src/components/claims/claim-details/components/ClaimType.tsx
palmenhq: no, type inference won't work and it technically becomes `undefined` instead of `false`. `undefined` is technically "falsey", so yes it'd work with the current implementation, but type safety is a thing and we should utilise it imo

PR #122, opened 2020-01-10
Feature/discount info hope

akashpatl:
Adding some more info under the account tab. This includes the total discount, total charge and what active discount code will be used for the next charge.

https://github.com/HedvigInsurance/account-service/pull/61
https://github.com/HedvigInsurance/hedgehog


src/components/chat/tabs/AccountTab.tsx
```diff
@@ -86,6 +91,25 @@ export const AccountTab: React.SFC<
           <h3>
             Balance (total): {formatMoneySE(data.member.account.totalBalance)}
           </h3>
+          <h3>Upcoming charge information:</h3>
+          <b4>
```
palmenhq: `b4`? Ive never heard of this tag, what does it do?

akashpatl: I naively assumed it would be a tag for body and the number for the size. Using the `<h2>` sets all the text to bold and quite large margins.

akashpatl: Stole it from MemberInformation.tsx

palmenhq: ah, i'm pretty sure there's no `b4` tag in the html spec and hence i think we should not use it. For next time i think we should address concerns like these before we merge :)

PR #128, opened 2020-01-27
Fix for create quotes when quotes exists

johantjellden:


src/components/chat/tabs/insurance-tab/create-quote.tsx
```diff
@@ -87,7 +87,7 @@ export const CreateQuote: React.FunctionComponent<{
     <Modal
       trigger={
         quotes
-          .map((quote) => quote.originatingProductId)
+          .map((quote) => (quote.state == "QUOTED" ? quote.originatingProductId : null)
```
palmenhq: Whe do we do this?
also please use `===` :)

johantjellden: Seems like we dont.

PR #131, opened 2020-02-04
Gatekeeper

palmenhq:
It might not be beautiful, but it does work. Coupled with https://github.com/HedvigInsurance/back-office/pull/238


src/server/auth.ts
```diff
@@ -0,0 +1,74 @@
...
+  { accessToken, refreshToken }: Tokens,
+) {
+  ctx.cookies.set('_hvg_at', accessToken, {
+    path: '/',
+    httpOnly: true,
+    expires: addMinutes(new Date(), 3 * 60),
+  })
+  ctx.cookies.set('_hvg_rt', refreshToken, {
+    path: '/',
+    httpOnly: true,
```
hedvigoscar: ❤️

src/server/config.ts
```diff
@@ -0,0 +1,30 @@
+const defaultGatekeeperHost =
+  process.env.NODE_ENV === 'production' ? null : 'https://id.dev.hedvigit.com'
```
vonElfvin: Shoult it he null here?

palmenhq: yes, this is only the default value. The actual value is set on [L27](https://github.com/HedvigInsurance/hedgehog/pull/131/files#diff-fdc54efe8bd543470b01b86f9eb46316R27)

PR #133, opened 2020-02-10
Make CI and UI great again ✊

palmenhq:
1. Adds CI to files that diff against master.
2. Adds structure for using one Hedvig-branded UI-lib by adding a button.
3. Uses the button in the quotes tab
4. Fixes the quotes tab in terms of typing and linting issues.

I say lets try with this solution and if it becomes too much work to fix stuff lets kill it off again.

👇 a sample of how we can work with a ui package

Screenshot 2020-02-10 at 10 10 50


src/components/chat/tabs/quotes/quote-modification.tsx
```diff
@@ -95,17 +92,18 @@ function createQuoteData({
...
       ...baseData,
-      ancillaryArea: ancillaryArea !== null ? parseInt(ancillaryArea) : null,
+      ancillaryArea:
+        ancillaryArea !== null ? parseInt(ancillaryArea, 10) : null,
       yearOfConstruction:
-        yearOfConstruction !== null ? parseInt(yearOfConstruction) : null,
+        yearOfConstruction !== null ? parseInt(yearOfConstruction, 10) : null,
       numberOfBathrooms:
-        numberOfBathrooms !== null ? parseInt(numberOfBathrooms) : null,
+        numberOfBathrooms !== null ? parseInt(numberOfBathrooms, 10) : null,
```
vonElfvin: What does 10 do here?

palmenhq: It's the radix, so for example `10` for decimals and `2` for the binary system. It defaults to 10 but the linter enforces us to add it. It's possible to turn off but it's very common to have i believe

PR #136, opened 2020-02-14
Patch the payment tab forms

palmenhq:
Fixes a weird issue where form fields suddenly jumped up and displayed horizontally instead of vertically. Much weird, very strange. Also upgrades some packages - while we're at it.


src/components/shared/inputs/TextField.tsx
```diff
@@ -1,6 +1,7 @@
 import { TextField as MuiTextField } from '@material-ui/core'
 import { FieldProps } from 'formik'
 import * as React from 'react'
+import styled from 'react-emotion'
```
johantjellden: Is this used below? Or do we import it for side effects?

palmenhq: Oj haha, no this is some kind of rest. I’ll remove it. But first - I will check why the compiler didn’t pick this up 🧐

palmenhq: Found it - i thought the `"noUnusedLocals"` was part of the `"strict"` mode but apparently not

PR #141, opened 2020-02-28
Reworked inventory

rasmusguterstam:
### **Redesigned and restructured "Itemizer"** 🤙

**In short:**
- Removes "item database", and disconnects the pricing service (for now)
- Implements Material-ui instead of Semantic for consistency 
- Restructured code and file structure (more consistent with other items under claim-details/components)

**Why:**
- Pricing service currently too slow and implementation off, needs to be reworked
- Inventory not intuitive/user-friendly
- The code & structure looked like s🤬🤬t, _slightly_ better now 🤪
- Enable a smooth future implementation of the "Item database" in the inventory as a supporting service (e.g. giving suggestions) rather than a standalone tool

image


rasmusguterstam:
Build fails while typechecking `src/components/claims/claim-details/index.tsx`, referring to error in `src/components/claims/claim-details/components/FileUpload.tsx`.

Got a feeling this isn't expected behavior? Or am I missing something? 😐

palmenhq:
@rasmusguterstam We've started incrementally fixing the code base by type checking and linting the differences against master. So this might have been wrong before but needs fixing now. Looking at the code there seems to be an inconsistency between `react-emotion` and the `dropzone` library. Looking at the error though, you could probably unset the `css` property and it should work, like so:`css={undefined}`

src/components/claims/claim-details/components/ClaimInventory.js
```diff
@@ -0,0 +1,321 @@
...
+  Table,
+  TableHead,
+  TableRow,
+  TableCell,
+  TableBody,
+  IconButton,
+  withStyles,
+} from '@material-ui/core'
+import MenuItem from '@material-ui/core/MenuItem'
+import DeleteForeverIcon from '@material-ui/icons/Delete'
```
palmenhq: can we not fetch this from `@material-ui/core`?

src/components/claims/claim-details/components/ClaimInventory.js
```diff
@@ -0,0 +1,321 @@
...
+import { formatMoney } from 'lib/intl'
+import { GET_INVENTORY, GET_CATEGORIES } from 'features/pricing/queries'
+import { ADD_ITEM, REMOVE_ITEM } from 'features/pricing/mutations'
```
palmenhq: I think we should move these in - either to this file - or if that is no good move them so  at least they're co-located with it

src/components/claims/claim-details/components/ClaimInventory.js
```diff
@@ -0,0 +1,321 @@
...
+                {formatMoney(
+                  'sv-SE',
+                  0,
+                )({
+                  amount: row.value,
+                  currency: 'SEK',
+                })}
+              </InventoryTableCell>
+              <InventoryTableCell>
+                <Mutation
```
palmenhq: i'd much rather use hooks here

src/components/claims/claim-details/components/ClaimInventory.js
```diff
@@ -0,0 +1,321 @@
...
+                  refetchQueries={() => {
+                    return [
+                      {
+                        query: GET_INVENTORY,
+                        variables: { claimId: this.props.claimId },
+                      },
+                    ]
+                  }}
+                >
+                  {(removeItem) => {
```
palmenhq: do we need any loading state here, or not really?

src/components/claims/claim-details/components/ClaimInventory.js
```diff
@@ -0,0 +1,321 @@
...
+    )
+  }
+
+  getNewItem = () => {
+    return {
+      inventoryItemId: null,
+      claimId: this.props.claimId,
+      itemName: this.state.itemName,
+      categoryName: this.state.itemCategory,
+      categoryId: '-1',
```
palmenhq: why not `null`? And why a number-string?

palmenhq:
also one more thing, really good that you killed off some code! But shall we also kill `features/pricing/index.js`?

src/components/claims/claim-details/components/ClaimInventory.tsx
```diff
@@ -0,0 +1,279 @@
...
+import { Paper } from '../../../shared/Paper'
+
+const GET_CATEGORIES = gql`
+  query Categories {
+    categories
+  }
+`
+
+const GET_INVENTORY = gql`
+  query Inventory($claimId: ID!) {
```
palmenhq: You could also make these `.graphql` files and get the hooks generated for you, just for the record :)

schema.graphql
```diff
@@ -8,17 +8,14 @@ type QueryType {
...
   ticket(id: ID!): Ticket
   getFullTicketHistory(id: ID!): TicketHistory
   tickets(resolved: Boolean): [Ticket!]!
   getAnswerSuggestion(question: String): [Suggestion!]!
   me: String
   inventory(claimId: ID!): [InventoryItem]!
-  filters(categoryId: String!): [FilterSuggestion!]!
-  inventoryItemFilters(inventoryItemId: String!): [FilterOutput]
+  categories: [String!]
+  itemSuggestions(query: String!): [SearchItem]!
```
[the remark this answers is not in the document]
rasmusguterstam: Good call, yes indeed! 😄

src/components/claims/claim-details/components/ClaimInventory/ClaimInventory.tsx
```diff
@@ -0,0 +1,391 @@
...
+
+  const getAggregatedResult = () => {
+    if (suggestionResult.length !== 0) {
+      const aggregatedResult = suggestionResult.itemSuggestions[0]
+
+      if (typeof aggregatedResult !== 'undefined') {
+        if (!aggregatedResult.url) {
+          return aggregatedResult
+        }
+      }
```
palmenhq: you could also use optional chaining like so: 
to save a few keystrokes ;)

```ts
if (!aggregatedResult?.url) {
  return aggregatedResult
}
```

And even if we don't use optional chaining it's usually better to not nest ifs. This could be rewritten like this (but optional chaining would be better imo):

```ts
if (typeof aggregatedResult === 'undefined' || !aggregatedResult.url) {
  return aggregatedResult
}
```

src/components/claims/claim-details/components/ClaimInventory/ClaimInventory.tsx
```diff
@@ -0,0 +1,391 @@
...
+    return { name: null, pricerunnerId: [], url: null }
+  }
+
+  const {
+    data: dataCategories,
+    loading: loadingCategories,
+    error: errorCategories,
+  } = useGetCategoriesQuery()
+
+  const { data = { inventory: undefined } } = useGetInventoryQuery({
```
palmenhq: I do not think the linter (or even the compiler?) will be happy about this, this would be equal to defining a variable called `undefined`, which is kinda claimed by JS already 😬 maybe you ment `data = { inventory = undefined } }` (but i do not understand why you would default something to `undefined` considering that's the default value)

src/components/claims/claim-details/components/ClaimInventory/ClaimInventory.tsx
```diff
@@ -0,0 +1,391 @@
...
+    variables: { claimId },
+  })
+
+  const { data: suggestionResult = [] } = useGetSuggestionsQuery({
+    variables: { query: itemName },
+  })
+
+  const {
+    data: smartSuggestion = { itemDetails: { category: null, price: null } },
```
palmenhq: same here, `null` is a reserved keyword. probably you want `=`, but again, why?

src/components/claims/claim-details/components/ClaimInventory/ClaimInventory.tsx
```diff
@@ -0,0 +1,391 @@
...
+                  refetchQueries: () => [
+                    {
+                      query: GetInventoryDocument,
+                      variables: { claimId },
+                    },
+                  ],
+                })
+                clearNewItem()
+              }
+            }}
```
palmenhq: i think we should do an early return here instead, so we don't get so many indentation levels:

```ts
if (formLooksGood) {
  return
}

await addInventoryItem({ ... })
...
```

src/components/claims/claim-details/components/ClaimInventory/ClaimInventory.tsx
```diff
@@ -0,0 +1,391 @@
...
+                      useSuggestion()
+                    }
+                  }}
+                  helperText={
+                    formLooksGood
+                      ? 'Press Return to add item ↩'
+                      : itemName !== '' &&
+                        suggestion.name &&
+                        suggestion.category &&
+                        itemName !== suggestion.name
```
palmenhq: ouff, could we make this a named function or something? It is really hard to read, especially in a nested ternary 😬

src/components/claims/claim-details/components/ClaimInventory/queries.graphql
```diff
@@ -0,0 +1,29 @@
+query GetInventory($claimId: ID!) {
```
palmenhq: Normally we place these queries in separate files, but maybe this is fine considering how tiny they are :) just for future reference

schema.graphql
```diff
@@ -90,85 +88,32 @@ type AutoLabel {
...
-type FilterSuggestion {
-  name: String!
-  items: [String!]!
-  others: [String!]!
+type SearchItem {
+  name: String
+  category: String
+  url: String
+  pricerunnerId: [String]
+  price: Float
```
palmenhq: should all of these really be nullable?

src/components/claims/claim-details/components/ClaimInventory/ClaimInventory.tsx
rasmusguterstam: Had no idea optional chaining is a thing - that looks really neat!! 😃

palmenhq: Its new since TS 3.7, which was released just a few months ago :)

PR #144, opened 2020-03-09
Feature/contracts tab

vonElfvin:
(A lot of the lines come from GraphQL codegen for the Contracts)
- Add a contracts tab with ability to activate pending agreements, also, terminating, changing termination date and reverting termination
- Add the following components to hedvig-ui: DateTimePicker, TextArea, ButtonsGroup, EnumDropdown
- Minor design update to Questions tab


shared/hedvig-ui/dropdown.tsx
```diff
@@ -0,0 +1,46 @@
...
+  placeholder: string
+  setValue: (value: any) => void
+}> = ({ enumToSelectFrom, placeholder, setValue }) => {
+  const dropdownOptions: DropdownItemProps[] = Object.values(
+    enumToSelectFrom,
+  ).map((value, index) => {
+    return {
+      key: index + 1,
+      value: value as string,
+      text: getTextFromEnumValue(value as string),
```
palmenhq: Shall we also give the user the option to just specify the text themselves? While this solution is really clever and nice i think for building quick things, it could be useful to also be able to choose the options yourself.

shared/hedvig-ui/dropdown.tsx
```diff
@@ -0,0 +1,46 @@
...
+})
+
+export const EnumDropdown: React.FunctionComponent<{
+  enumToSelectFrom: any
+  placeholder: string
+  setValue: (value: any) => void
+}> = ({ enumToSelectFrom, placeholder, setValue }) => {
+  const dropdownOptions: DropdownItemProps[] = Object.values(
+    enumToSelectFrom,
+  ).map((value, index) => {
```
palmenhq: what'll happen if you haven't specified string values for the enum? Then the `Object#values` will return a list of numbers, which'll be problematic with this implementation

shared/hedvig-ui/text-area.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+export const TextAreaThatGrows: React.FunctionComponent = () => {
+  const [text, setText] = React.useState('')
+  return (
+    <>
+      <TextArea
+        placeholder={'Write your life story here...'}
+        setText={setText}
+      />
+      <p>
+        <b>Written text:</b> {text}
```
palmenhq: `strong`, please :)

shared/hedvig-ui/text-area.tsx
```diff
@@ -0,0 +1,23 @@
...
+  marginTop: '1rem',
+  width: '100%',
+})
+
+export const TextArea: React.FunctionComponent<{
+  placeholder: string
+  setText: (value: string) => void
+}> = ({ placeholder, setText }) => {
+  return (
+    <StyledForm>
```
palmenhq: what if this text area is part of a different context than standalone? I think we shouldn't wrap it in a form by default. Or, rather, not at all.

src/api/generated/graphql.tsx
```diff
@@ -1,33 +1,33 @@
+import gql from 'graphql-tag'
```
palmenhq: I believe tslint will fail this file _hard_, but we can fix it with `yarn lint --fix`

src/components/chat/tabs/contract-tab/contract-item.tsx
```diff
@@ -0,0 +1,36 @@
...
+          <ThirdLevelHeadline>Master Inception</ThirdLevelHeadline>
+          <MasterInception contract={contract} />
+        </Card>
+        <Card span={2}>
+          <ThirdLevelHeadline>Termination Date</ThirdLevelHeadline>
+          <TerminationDate contract={contract} />
+        </Card>
+        <Card span={2}>
+          <span>
+            <b>Contract id:</b> {contract.id}
```
palmenhq: `strong` ;)

src/components/chat/tabs/contract-tab/master-inception.tsx
```diff
@@ -0,0 +1,60 @@
+import { Contract } from 'api/generated/graphql'
+import { Button, ButtonsGroup } from 'hedvig-ui/button'
+import { DateTimePicker } from 'hedvig-ui/date-time-picker'
+import { FourthLevelHeadline, Paragraph } from 'hedvig-ui/typography'
+import {
+  activateContractOptions,
+  useActivateContract,
+} from 'hooks/use-activate-contract'
+import moment from 'moment'
```
palmenhq: `date-fns` ;)

src/components/chat/tabs/contract-tab/termination-date.tsx
```diff
@@ -0,0 +1,155 @@
...
+} from 'hooks/use-change-termination-date'
+import {
+  revertTerminationOptions,
+  useRevertTermination,
+} from 'hooks/use-revert-termination'
+import {
+  terminateContractOptions,
+  useTerminateContract,
+} from 'hooks/use-terminate-contract'
+import moment, { Moment } from 'moment'
```
palmenhq: you know the drill

src/components/chat/tabs/contract-tab/termination-date.tsx
```diff
@@ -0,0 +1,155 @@
...
+                  onClick={() =>
+                    revertTermination(revertTerminationOptions(contract)).then(
+                      reset,
+                    )
+                  }
+                >
+                  Are you sure?
+                </Button>
+              )}
+            </ButtonsGroup>
```
palmenhq: I know we've done this (swapping the buttons for confirmation) in the past, but from my understanding its been confusing for the users. An _acceptable_ solution imo, that gives way better UX is using the [`window.confirm`](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) dialog.

src/components/chat/tabs/index.js
```diff
@@ -1,19 +1,20 @@
...
 import ClaimsTab from 'components/chat/tabs/ClaimsTab'
-import { MemberDebtComponent } from 'components/chat/tabs/DebtTab'
+import {MemberDebtComponent} from 'components/chat/tabs/DebtTab'
 import DetailsTab from 'components/chat/tabs/DetailsTab'
 import MemberFile from 'components/chat/tabs/FileTab'
 import InsuranceListTab from 'components/chat/tabs/InsuranceListTab'
 import InsuranceTab from 'components/chat/tabs/InsuranceTab'
 import PaymentsTab from 'components/chat/tabs/PaymentsTab'
-import { Quotes } from 'components/chat/tabs/quotes'
+import {Quotes} from 'components/chat/tabs/quote-tab'
```
palmenhq: can we run prettier pls :)

src/components/chat/tabs/quote-tab/quote-activation.tsx
```diff
@@ -6,19 +6,23 @@ import { Button } from 'hedvig-ui/button'
 import * as React from 'react'
 import { Checkbox } from 'semantic-ui-react'
 import { noopFunction } from 'utils'
+import { QUOTES_QUERY } from '../../../../hooks/use-quotes'
```
palmenhq: i think we could shorten this import

src/components/chat/tabs/quote-tab/quote-modification.tsx
```diff
@@ -20,8 +20,8 @@ import styled from 'react-emotion'
 import { Checkbox, Dropdown, Input as SuiInput } from 'semantic-ui-react'
 import { noopFunction } from 'utils'
 import * as uuid from 'uuid/v4'
+import { QUOTES_QUERY } from '../../../../hooks/use-quotes'
```
palmenhq: same, this could be shortened :)

src/components/chat/tabs/quote-tab/quote-modification.tsx
```diff
@@ -367,7 +368,7 @@ const ExtraBuildingEditor: React.FC<{
         <strong>Extra buildings</strong>
       </div>
       {extraBuildings.map((extraBuilding, i) => (
-        <ExtraBuildingWrapper key={extraBuilding?.id}>
+        <ExtraBuildingWrapper key={extraBuilding?.id ?? uuid()}>
```
palmenhq: if `extraBuldiing?.id` is empty it'll get a new key every render with this solution. That means that, for example, on every key stroke this item it'll be removed from the DOM and added again, which might make the page slow, but also the user will have to focus on the field they're typing in on every key stroke.

src/components/questions/questions-list/QuestionsList.js
```diff
@@ -8,8 +8,8 @@ import { isMemberIdEven } from 'utils/member'
 const ListContainer = styled.div`
   display: flex;
   flex-direction: column;
-  width: 700px;
-  margin: 0 auto 50px;
+  width: 100%;
```
palmenhq: nice

src/components/shared/inputs/DateInput.tsx
```diff
@@ -24,8 +24,9 @@ const WidgetContainer = styled('div')({
 })
 
 const DatePickerContainer = styled('div')({
-  fontFamily: 'Circular Std Book !important',
+  fontFamily: 'Circular Std Book !important, Arial',
```
palmenhq: i think `!important` must go last for this to work. Also, can we maybe default to `sans-serif` rather than `Arial`? I'm not sure arial comes with every computer.

src/hooks/use-activate-contract.tsx
```diff
@@ -0,0 +1,40 @@
+import { useMutation } from '@apollo/react-hooks'
```
palmenhq: why have we placed these hooks in a global package? Its only used inside the `contract` tab right?

src/hooks/use-activate-contract.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { gql } from 'apollo-boost'
+import { Moment } from 'moment'
```
palmenhq: ;)

src/hooks/use-activate-contract.tsx
```diff
@@ -0,0 +1,40 @@
...
+const ACTIVATE_PENDING_AGREMENT = gql`
+  mutation ActivatePendingAgreement($request: ActivatePendingAgreementInput) {
+    activatePendingAgreement(request: $request) {
+      id
+    }
+  }
+`
+
+export const useActivateContract = (contract: Contract) => {
+  const [activateContract] = useMutation(ACTIVATE_PENDING_AGREMENT, {
```
palmenhq: maybe we could return `useMutation` directly? In case you wan't access to `error` or `loading`, that are present in `[1]` of the returned tuple

src/hooks/use-change-termination-date.tsx
```diff
@@ -0,0 +1,57 @@
+import { ExecutionResult } from '@apollo/react-common'
+import { useMutation } from '@apollo/react-hooks'
+import { ChangeTerminationDateInput, Contract } from 'api/generated/graphql'
+import { gql } from 'apollo-boost'
+import { Moment } from 'moment'
```
palmenhq: ⏰

src/hooks/use-change-termination-date.tsx
```diff
@@ -0,0 +1,57 @@
...
+      id
+    }
+  }
+`
+
+export const useChangeTerminationDate = (
+  contract: Contract,
+  onCompleted: (value?: any) => void = noopFunction,
+): [
+  (ChangeTerminationDateInput) => Promise<ExecutionResult<Contract>>,
```
palmenhq: don't we need a variable name first? like so: `(input: ChangeTerminationDateInput) => Promise<ExecutionResult<Contract>>,`?

src/hooks/use-contracts.tsx
```diff
@@ -0,0 +1,106 @@
+import { useQuery } from '@apollo/react-hooks'
+import { gql } from 'apollo-boost'
+import { Contract, Member } from '../api/generated/graphql'
+
+export const CONTRACTS_QUERY = gql`
```
palmenhq: we could generate hooks for these queries by placing them in a `.graphql` file. That way, if you run `yarn graphq:gen`, you'll get the hooks for free in `generated/graphql.tsx`

src/hooks/use-terminate-contract.tsx
```diff
@@ -0,0 +1,62 @@
...
+`
+
+export const useTerminateContract = (
+  contract: Contract,
+  onCompleted: (value?: any) => void = noopFunction,
+): [
+  (ChangeTerminationDateInput) => Promise<ExecutionResult<Contract>>,
+  boolean,
+] => {
+  const [terminateContract, { loading }] = useMutation(TERMINATE_CONTRACT, {
```
palmenhq: same here, why not just return `useMutation` directly?

shared/hedvig-ui/button.tsx
```diff
@@ -38,14 +58,22 @@ export const buttonSizeMap: Record<
...
+    float,
     display: 'inline-flex',
     alignItems: 'center',
     justifyContent: 'center',
     whiteSpace: 'nowrap',
     fontSize: buttonSizeMap[size].fontSize,
     padding: buttonSizeMap[size].padding,
-    width: fullWidth ? '100%' : 'auto',
+    margin: '0.5rem',
+    width: fullWidth ? 'calc(100% - 1rem)' : 'auto',
```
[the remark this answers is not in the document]
vonElfvin: Yea this should be moved to the ButtonsGround component's css. I added it to create some gap between the buttons but put it on the wrong location 😇

schema.graphql
```diff
@@ -23,55 +23,203 @@ type QueryType {
...
-  ): AutoLabel
+  updateClaimState(id: ID! state: ClaimState!): Claim
+  createClaim(memberId: ID! date: LocalDateTime! source: ClaimSource!): ID
+  addClaimNote(id: ID! note: ClaimNoteInput!): Claim
+  createClaimPayment(id: ID! payment: ClaimPaymentInput!): Claim
+  setClaimType(id: ID! type: ClaimTypes!): Claim
+  setClaimInformation(id: ID! information: ClaimInformationInput!): Claim
+  updateReserve(id: ID! amount: MonetaryAmount!): Claim
+  setCoveringEmployee(id: ID! coveringEmployee: Boolean!): Claim
+  createTicket (ticket: TicketInput): ID
```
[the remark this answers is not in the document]
vonElfvin: This is copied from back-office haha, but yes I can copy them back and forth I guess 🛩

schema.graphql
```diff
@@ -889,19 +1042,20 @@ enum TicketType {
...
-# A String-representation of `java.time.LocalTime`,
+# A String-representation of `java.time.LocalTime`
 scalar LocalTime
 # A String-representation of `java.time.LocalDateTIme`. ex: `"2018-06-11T20:08:30.123456"`
 scalar LocalDateTime
 # A String-representation of `java.time.Instant`. ex: `"2018-06-11T20:08:30.123456"`
 scalar Instant
-# A String-representation of `java.net.URL`, ex: "https://www.google.com/"
+# A String-representation of `java.net.URL` ex: "https://www.google.com/"
 scalar URL
```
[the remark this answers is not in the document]
vonElfvin: Did a replace on ',' before, guess there were some in the comments haha

shared/hedvig-ui/button.tsx
```diff
@@ -54,10 +82,16 @@ export const Button = styled('button')<ButtonProps>(
     boxShadow: 'none !important',
     borderRadius: 5,
     cursor: 'pointer',
-
     '&:hover, &:focus': {
+      outline: 'none',
+      opacity: '0.8',
```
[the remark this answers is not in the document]
vonElfvin: Added highlighted prop

shared/hedvig-ui/dropdown.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+enum Taste {
+  Sweet = 'SWEET',
+  Sour = 'SOUR',
+  TheHedvigFavorite = 'THE_HEDVIG_FAVORITE',
+}
+
+export const DropdownWithEnum: React.FunctionComponent = () => {
+  const [terminationReason, setTerminationReason] = React.useState(null)
```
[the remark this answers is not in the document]
vonElfvin: Oh yea that was a mistake haha

shared/hedvig-ui/dropdown.tsx
vonElfvin: I would say we cross that bridge when we get there :)

src/components/chat/tabs/quote-tab/quote-modification.tsx
vonElfvin: It was whining when key couldn't be null, changed it to `?? undefined` now so it stopped whining but also doesn't generate a new id every time

shared/hedvig-ui/dropdown.tsx
palmenhq: Could we not just use `Object#keys` instead? or not really

shared/hedvig-ui/dropdown.tsx
vonElfvin: Even though I see what you mean I think its up for the future developer who want's to do that to update the generic component? Otherwise it feels like we are preoptimizing. It would be easy to change that so I think it can wait until we reach that point.

shared/hedvig-ui/dropdown.tsx
vonElfvin: We could but then it would be a lot harder to make the text look nice

vonElfvin: Making "NO_FEEDBACK" look nice is easier than making "NoFeedback" look nice since you can split on '_'. We could probably solve using the key one, I just didn't figure it out. I'd say it's more rare to use numbers as well given we never do that anywhere in the backend afaik.

src/components/chat/tabs/contract-tab/termination-date.tsx
```diff
@@ -0,0 +1,222 @@
...
+                      .then(() => {
+                        showNotification({
+                          type: 'olive',
+                          header: 'Termination date changed',
+                          message: 'Successfully changed termination date.',
+                        })
+                        reset()
+                      })
+                      .catch((error) => {
+                        showNotification({
```
palmenhq: Just a thought in these catch blocks, shall we rethrow the error, or at least write it to the console, or capture it with sentry, or something? Just to not "swallow" it? (applies to all these catch blocks i guess)

src/components/chat/tabs/index.js
```diff
@@ -7,7 +7,8 @@ import MemberFile from 'components/chat/tabs/FileTab'
 import InsuranceListTab from 'components/chat/tabs/InsuranceListTab'
 import InsuranceTab from 'components/chat/tabs/InsuranceTab'
 import PaymentsTab from 'components/chat/tabs/PaymentsTab'
-import { Quotes } from 'components/chat/tabs/quotes'
+import { Quotes } from 'components/chat/tabs/quote-tab'
+import { Contracts } from 'components/chat/tabs//contract-tab'
```
palmenhq: `//`?

src/utils/member.tsx
```diff
@@ -95,7 +95,7 @@ export const getMemberIdColor = (memberId: string) => {
   }
 
   if (isMemberIdEven(memberId)) {
-    return colorsV2.midnight500
+    return colorsV2.coral500
```
palmenhq: the actual colors 😂 right?

shared/hedvig-ui/button.tsx
```diff
@@ -1,22 +1,66 @@
...
+  width: '100%',
+  display: 'flex',
+  flexDirection: 'row',
+  paddingTop: '1rem',
+  '&& button:not(:last-of-type)': {
+    marginRight: '0.5rem',
+  },
+  '&& button:not(:first-of-type)': {
+    marginLeft: '0.5rem',
+  },
```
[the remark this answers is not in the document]
vonElfvin: I have never written code like this before but I guess that would work as well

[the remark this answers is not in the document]
vonElfvin: I don't think so since its not having marginRight on the last one and marginLeft on the first one. If we have that code then it would make it tricky to fix either case.

package.json
```diff
@@ -23,6 +23,8 @@
     "constate": "^0.8.2",
     "copy-webpack-plugin": "^4.5.2",
     "core-js": "2",
+    "create-react-context": "^0.3.0",
```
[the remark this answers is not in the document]
vonElfvin: Another package needs it and is unable to add it itself for unknown reason.

src/components/chat/tabs/index.js
vonElfvin: Weird that it even works

src/components/chat/tabs/contract-tab/termination-date.tsx
vonElfvin: Throwing them again after the notification

src/components/chat/tabs/contract-tab/index.tsx
```diff
@@ -0,0 +1,34 @@
+import { useContracts } from 'graphql/use-contracts'
+import * as React from 'react'
+import styled from 'react-emotion'
+import { Icon } from 'semantic-ui-react'
+import { Button } from '../../../../../shared/hedvig-ui/button'
+import { Spacing } from '../../../../../shared/hedvig-ui/spacing'
+import { MainHeadline } from '../../../../../shared/hedvig-ui/typography'
```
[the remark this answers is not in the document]
vonElfvin: Fixed it

shared/hedvig-ui/button.tsx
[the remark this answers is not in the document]
palmenhq: Sorry if that was clear haha, what i mean is the `&&`s. I don't think we need them. I understand you need separate clauses for `first-of-type` and `last-of-type`

package.json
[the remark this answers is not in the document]
palmenhq: so strange..

src/components/chat/tabs/contract-tab/contract-item.tsx
```diff
@@ -0,0 +1,33 @@
+import { Card, CardsWrapper } from 'hedvig-ui/card'
+import { SecondLevelHeadline, ThirdLevelHeadline } from 'hedvig-ui/typography'
+import * as React from 'react'
+import { Contract } from 'src/api/generated/graphql'
```
palmenhq: does this even work 🤨

PR #145, opened 2020-03-12
Feature/make chat accessible in claims

gustafvh:


src/components/chat/messages/MessagesList.js
```diff
@@ -27,6 +27,17 @@ export default class MessagesList extends React.Component {
   }
 
   componentDidUpdate() {
+    this.theRealScroller()
+    /* eslint-enable */
```
palmenhq: ?

src/components/chat/tabs/ChatTab.js
```diff
@@ -201,22 +287,24 @@ export default class ChatTab extends React.Component {
...
-      onClick={() =>
-        props.ctx.setState({
-          visible: !props.ctx.state.visible,
-          manualChange: true,
-        })
-      }
-    />
-  </ChatHeaderStyle>
-)
+class ChatHeader extends React.Component {
```
palmenhq: whats the reason for turning this into a class component?

src/components/chat/tabs/ChatTab.js
```diff
@@ -53,11 +56,25 @@ export default class ChatTab extends React.Component {
     this.state = {
```
palmenhq: just an idea - since this is technically not really a tab, but rathe a.... Pane? Maybe we should find a better name?

src/components/chat/tabs/ChatTab.js
```diff
@@ -166,13 +252,13 @@ export default class ChatTab extends React.Component {
...
     const lastNonMemberIndex = fromIds
-      .map((id) => id === +this.props.match.params.id)
+      .map((id) => id === +this.props.match.params.memberId)
       .lastIndexOf(false)
 
     const lastMemberMessagesArray = messages.filter(
       (message, index) =>
         ['free.chat.message', 'free.chat.from.bo', 'free.chat.start'].includes(
-          message.id,
+          message.memberId,
```
palmenhq: this seems suspicious

src/components/chat/tabs/ChatTab.js
palmenhq: And on that note, maybe we should try to extract the file location from `chat`, considering it's no longer part of the member page?

PR #147, opened 2020-03-18
Feature/norwegian tariff

vonElfvin:
Adds a simple tool for passing the data to the price engine service to create the norwegian price engine and also add a bunch of postal codes for the calculation. This will only be ran once and you cannot run it more. And easily can be built upon in the future when we want to give the ability to update the tariff. There is tons of validations being performed in the back-end to ensure the data is sent in a correct format.

The inputs include:
- Base factors for the different insurance products (Base factor for Fire, Leakage etc.)
- Factors for the different insurance products (Factor for Age in Fire, Age in Leakage, ..., Number of people Fire, Number of people Leakage etc.)
- Postal codes to get centrality group and economy of municipality from.


src/features/tools/index.tsx
```diff
@@ -12,19 +12,24 @@ const CardLink = Card.withComponent(Link)
...
     </CardLink>
 
-    <CardLink to="/tools/perils-editor" span={3}>
+    <CardLink to="/tools/perils-editor" span={4}>
       <Icon>📝</Icon>
       Perils editor
     </CardLink>
+
+    <CardLink to="/tools/norwegian-tariff-editor" span={4}>
+      <Icon>🇳🇴</Icon>
```
palmenhq: 🇳🇴💵 ? :D

src/features/tools/norwegian-tariff-editor/factor-editor.tsx
```diff
@@ -0,0 +1,33 @@
...
+  factorString,
+  setFactorString,
+}) => {
+  return (
+    <>
+      <SecondLevelHeadline>{factorName}</SecondLevelHeadline>
+      <CardsWrapper>
+        <Card span={2}>
+          <TextArea
+            placeholder={`Add factors of ${factorName} from excel (Including the name of the factor, its keys on the left and the columns: 'Fire', 'Leakage', 'Other', 'All risk' and 'Travel' OBS: Do not include 'Special object')...`}
```
palmenhq: should this maybe be an instruction aside from the text area, instead of a placeholder that disappears when you start typing?

src/features/tools/norwegian-tariff-editor/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+        <Button
+          fullWidth
+          variation={'primary'}
+          disabled={loading}
+          onClick={() => {
+            if (
+              window.confirm(
+                'Are you sure you want to create the price engine?',
+              )
+            ) {
```
palmenhq: maybe we could negate this if and make an early return instead? Just to not get so many indentation levels

src/features/tools/norwegian-tariff-editor/postal-codes-editor.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const [
+    addNorwegianPostalCodes,
+    { loading },
+  ] = useAddNorwegainPostalCodesMutation()
+
+  return (
+    <Spacing top>
+      <TextArea
+        placeholder={
+          'Add postal codes columns from excel (Columns: "Postnummer", "Poststed", "Municipality name", "Disposable income" and "Centrality group")'
```
palmenhq: Same, i think this should be an instruction aside from the text field

src/utils/enum.ts
```diff
@@ -0,0 +1,12 @@
...
+    .toLowerCase()
+    .split('_')
+    .map((word, index) => {
+      if (index === 0 || word === 'hedvig') {
+        return word.charAt(0).toUpperCase() + word.slice(1)
+      }
+      return word
+    })
+    .join(' ')
+}
```
palmenhq: While moving this is clever, i think we should let it reside within `hedvig-ui` in order for it to be possible to break that package out to a separate repository, if that's something we'd want to do.

src/features/tools/norwegian-tariff-editor/factor-editor.tsx
```diff
@@ -0,0 +1,33 @@
...
+            setText={setFactorString}
```
palmenhq: a best practice is to set `value` of the text field you're editing also.

src/features/tools/norwegian-tariff-editor/factor-editor.tsx
```diff
@@ -0,0 +1,33 @@
...
+          />
+        </Card>
+        <Card span={2}>
+          <Paragraph>{factorString}</Paragraph>
```
palmenhq: Not sure i understand this ui exactly, i can. see already what i'm entering in the textfield already, right? or what's the purpose of printing it again? If its because the text field is too tiny i think rather we should expand the text field

src/features/tools/norwegian-tariff-editor/factor-editor.tsx
vonElfvin: Given this is just for me more or less I think this is sufficient for now. The UX is not crisp 😄

src/features/tools/norwegian-tariff-editor/index.tsx
vonElfvin: Smart

src/features/tools/norwegian-tariff-editor/factor-editor.tsx
vonElfvin: TextArea has an internal value + setValue state, this is just tapping into that logic. Where in this case I want a more complex setter than a regular value.

src/features/tools/norwegian-tariff-editor/factor-editor.tsx
vonElfvin: Yea this was mostly for debugging purposes. I can remove it.

vonElfvin: Yea this was mostly for debugging purposes. I can remove it.

vonElfvin:
> Over all really awesome!
> 
> Just out of curiosity, what happens if you use this tool multiple times? Do we have any "undo" logic? Or like whats the use case for it, more than applying the initial tariff?

If you try again nothing happens, there is no undo logic for now except for cleaning the database manually. I don't think we should undo it regardless. 

There is no other application than to apply the initial tariff.

src/features/tools/norwegian-tariff-editor/factor-editor.tsx
vonElfvin: For example before I replaced all ' \t' with '\\t', ',' with '.'and stuff like that, I later realised I didn't need that. However, with having an external "setText" it was super easy to do extra stuff while not impacting the text that is actually written in the form.

src/utils/enum.ts
vonElfvin: Moved it back and added some custom logic for my use case

src/features/tools/index.tsx
vonElfvin: Having two icons looked weird and don't feel like I have the time to figure out how to fix it, I did change it to an airplane though since the engine is called "Gripen" 🛩

src/features/tools/norwegian-tariff-editor/factor-editor.tsx
palmenhq: Aha! But maybe you could do that late rather than early? So you do it when you need to?

src/features/tools/norwegian-tariff-editor/factor-editor.tsx
palmenhq: I think we should fix it though? If someone wants to take inspo from this code in the future if not

PR #153, opened 2020-03-26
Feature flag

vonElfvin:
Inject ContractMarketInfo to each member tab, show placeholder for pages that are sweden only (so far)


vonElfvin:
> I'm not sure what is better, doing ["duck typing"](https://en.wikipedia.org/wiki/Duck_typing) for whether to display features (something in the lines of _"if it has a direct debit account set up, then direct debit is available"_), or checking whether that feature is available for specific markets or not. I kinda like the former more, because if we start doing market-specific checks for all things that strategy will probably stick around for a while and it can quickly become a nasty mess i think. Especially if we introduce more markets. In some cases maybe its ok, but i think these particular examples would benefit from the duck-approach rather. Unless its temporary. Whats your thoughts?

I agree with you completely, I will explore "per feature" if duck typing is available, otherwise, as you said, this is a rather temporary as a "child safety" precaution so nothing unexpected happens via IEX poking around in H.OPE. until we've fixed all features completely.

palmenhq:
@vonElfvin haha, i understand. Will we fix it though? Can we if not, leave some comments or something, saying something in the lines of `FIXME, we should use check for availability rather than market, don't do this` or so?

vonElfvin:
> @vonElfvin haha, i understand. Will we fix it though? Can we if not, leave some comments or something, saying something in the lines of `FIXME, we should use check for availability rather than market, don't do this` or so?

Fixing these stuff are what we are going to do now to enable Norway. Now we just need to make the pages accessible without crashing to make it easier to develop. I added FIXME comments now that we can take off one by one.

PR #154, opened 2020-03-27
add icons to perils editor

AlexSJ1990:


src/features/tools/perils-editor/index.tsx
```diff
@@ -243,6 +244,17 @@ export const PerilsEditorComponent: React.FC<WithShowNotification> = ({
                   </Coverage>
                 </CoverageWrapper>
 
+                <span>Add Icon</span>
+                <Dropdown
+                  placeholder={peril.icon != null ? peril.icon : 'Icon name'}
```
palmenhq: There's no case where you shouldn't use triple equals :) `!==`/`===` always always always. BUT in this particular case you could use the new and fancy null coalescing operator: `peril.icon ?? 'Icon name'`

src/features/tools/perils-editor/index.tsx
```diff
@@ -243,6 +245,17 @@ export const PerilsEditorComponent: React.FC<WithShowNotification> = ({
                   </Coverage>
                 </CoverageWrapper>
 
+                <span>Add Icon</span>
+                <Dropdown
+                  placeholder={peril.iconName ?? 'Icon name'}
```
palmenhq: should the `peril.iconName` be the `value` rather? or no?

PR #168, opened 2020-04-29
Use GraphQL for Questions tab

vonElfvin:


src/components/tickets/ticket/body/message-response.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { Mutation } from 'react-apollo'
 import { connect } from 'react-redux'
 import { Button } from 'semantic-ui-react'
-import { QUESTION_IS_DONE } from '../../../../features/taskmanager/queries'
+import { QUESTION_IS_DONE } from 'features/taskmanager/queries'
 import actions from '../../../../store/actions/index'
```
AlexSJ1990, line 6: this one too - but v minor

src/components/questions/questions-list/AnswerForm.js
```diff
@@ -18,71 +27,70 @@ const FormTextArea = styled(Form.Field)`
...
+            placeholder="Answer text..."
+            onChange={answerChangeHandler}
+            value={answer}
+          />
+          <div>
+            <Button
+              style={{ marginBottom: '3px' }}
+              content="Open Chat"
+              onClick={() => chatRedirectClick()}
+            />
```
palmenhq, line 78: I understand probably it was like this before but imo thi should be a `Link` not a `Button` with `redirect` :) boy scout rule and all that

src/features/taskmanager/queries.ts
```diff
@@ -83,6 +83,6 @@ export const CHANGE_REMINDER = gql`
 
 export const QUESTION_IS_DONE = gql`
   mutation QuestionIsDone($memberId: ID!) {
```
palmenhq, line 85: What about renaming these to the same thing?

src/graphql/use-mark-question-as-done.tsx
```diff
@@ -0,0 +1,27 @@
...
+import {
+  GetQuestionsGroupsDocument,
+  MarkQuestionAsDoneMutation,
+  MarkQuestionAsDoneMutationVariables,
+  useMarkQuestionAsDoneMutation,
+} from 'api/generated/graphql'
+
+export const useMarkQuestionAsDone = () => useMarkQuestionAsDoneMutation()
+
+export const markQuestionAsDoneOptions = (
```
palmenhq, line 11: maybe we should prefix this with `get`? To clarify the action is not marking but rather getting the options

vonElfvin:
Had to do this from scratch again, which is already out in production so

PR #169, opened 2020-05-04
add norwegian product types to modify quote

AlexSJ1990:


src/components/member/tabs/quote-tab/quote-modification.tsx
```diff
@@ -70,6 +71,19 @@ const UPDATE_QUOTE_MUTATION = gql`
...
+  if (quote.productType === 'APARTMENT') {
+    return (quote.data as ApartmentQuoteData).subType!.toString()
+  }
+  if (quote.productType === 'HOUSE') {
+    return 'HOUSE'
+  }
+  if (quote.productType === 'TRAVEL') {
+    return 'TRAVEL'
+  }
+}
```
vonElfvin, line 85: What about home contents?

src/components/member/tabs/quote-tab/quote-modification.tsx
```diff
@@ -114,6 +128,18 @@ const createQuoteData = ({
       ),
       isSubleted: isSubleted ?? false,
     }
+  }
+  if ((productType ?? quote.productType) === 'HOME_CONTENT') {
```
vonElfvin, line 132: Use `getProductTypeValue` here?

vonElfvin: Alternatively create a helper function like `isHomeContent(quote)`

src/components/member/tabs/quote-tab/quote-modification.tsx
```diff
@@ -176,6 +202,29 @@ export const QuoteModification: React.FC<{
...
+      }}
+      options={[
+        { text: 'Apartment (rent)', value: 'RENT' },
+        { text: 'Apartment (brf)', value: 'BRF' },
+        { text: 'Apartment (student rent)', value: 'STUDENT_RENT' },
+        { text: 'Apartment (student brf)', value: 'STUDENT_BRF' },
+        { text: 'House', value: 'HOUSE' },
+        { text: 'Norwegian Home Content', value: 'HOME_CONTENT' },
+        { text: 'Norwegian Travel', value: 'TRAVEL' },
+      ]}
```
vonElfvin, line 224: Ah fuck, when we are able to create contracts from hope this should be different for the different kinds of contracts.

src/components/member/tabs/quote-tab/quote-modification.tsx
```diff
@@ -228,40 +277,28 @@ export const QuoteModification: React.FC<{
...
-
-        {getNumberInput('livingSpace', 'Living space (m2)')}
-        {getNumberInput('householdSize', 'Household size (# of people)')}
-      </InputGroup>
+      {(formState.productType ?? quote.productType) === 'TRAVEL' && (
+        <InputGroup>
+          <Label htmlFor={`producttype-${quote.id}`}>Product type</Label>
+          {getProductTypeDropdown()}
+          {getNumberInput('householdSize', 'Household size (# of people)')}
+        </InputGroup>
```
vonElfvin, line 285: Why would this one be travel only?

src/components/member/tabs/quote-tab/quote-modification.tsx
```diff
@@ -176,6 +202,29 @@ export const QuoteModification: React.FC<{
...
+        setFormState({ ...formState, productType: data.value as string })
+      }}
+      options={[
+        { text: 'Apartment (rent)', value: 'RENT' },
+        { text: 'Apartment (brf)', value: 'BRF' },
+        { text: 'Apartment (student rent)', value: 'STUDENT_RENT' },
+        { text: 'Apartment (student brf)', value: 'STUDENT_BRF' },
+        { text: 'House', value: 'HOUSE' },
+        { text: 'Norwegian Home Content', value: 'HOME_CONTENT' },
+        { text: 'Norwegian Travel', value: 'TRAVEL' },
```
vonElfvin, line 223: Just to make sure, I think I implemented it as "isYouth" was extracted from the type instead of actually having a checkbox "isYouth". So if we only have "HOME_CONTENT" here then we should also have that checkbox. Or if we do not have that checkbox we should have "HOME_CONTENT_YOUTH" as an option.

src/components/member/tabs/quote-tab/quote-modification.tsx
AlexSJ1990: yes this is true

src/components/member/tabs/quote-tab/quote-modification.tsx
AlexSJ1990: should we also have TRAVEL_YOUTH as an option then also?

src/api/generated/graphql.tsx
```diff
@@ -1052,6 +1055,7 @@ export type NorwegianTravelQuoteData = IQuoteData & {
   firstName?: Maybe<Scalars['String']>
   lastName?: Maybe<Scalars['String']>
   householdSize?: Maybe<Scalars['Int']>
+  isYouth?: Maybe<Scalars['Boolean']>
```
vonElfvin, line 1058: This should probably not be here right? Since you have the `NorwegianTravelLineOfBusiness` above

src/components/member/tabs/quote-tab/quote-modification.tsx
```diff
@@ -70,6 +78,93 @@ const UPDATE_QUOTE_MUTATION = gql`
...
+  if (isSwedishHouse(quote)) {
+    return QuoteProductType.House
+  }
+  if (isNorwegianHomeContent(quote)) {
+    return QuoteProductType.HomeContent
+  }
+  if (isNorwegianTravel(quote)) {
+    return QuoteProductType.Travel
+  }
+}
```
vonElfvin, line 133: throw here

src/components/member/tabs/quote-tab/quote-modification.tsx
```diff
@@ -70,6 +78,93 @@ const UPDATE_QUOTE_MUTATION = gql`
...
+
+const isNorwegianHomeContentFormStateProductSubType = (
+  productType: string | null,
+): boolean => {
+  if (productType === null) {
+    return false
+  }
+  const subTypes: string[] = [
+    NorwegianHomeContentType.Own.valueOf(),
+    'HOME_CONTENT_RENT',
```
vonElfvin, line 161: Why not 
```suggestion
   NorwegianHomeContentType.Own.valueOf(),
```

src/components/member/tabs/quote-tab/quote-modification.tsx
```diff
@@ -176,6 +300,41 @@ export const QuoteModification: React.FC<{
...
+            text: 'Norwegian Home Content (youth own)',
+            value: 'YOUTH_OWN',
+          },
+          {
+            text: 'Norwegian Home Content (youth rent)',
+            value: 'YOUTH_RENT',
+          },
+          { text: 'Norwegian Travel', value: 'REGULAR' },
+          { text: 'Norwegian Travel (youth)', value: 'YOUTH' },
+        ]}
```
vonElfvin, line 333: This one could probably be split up depending on the market actually. We allow switching  from house to apartment so IEX can create a qutoe but going from norway to sweden sounds unnecessary. Could probably use `contractMarketInfo.market`

src/graphql/get-quotes.graphql
```diff
@@ -38,6 +38,20 @@ query GetQuotes($memberId: ID!) {
...
           isSubleted
         }
+
+        ... on NorwegianHomeContentQuoteData {
+          street
+          zipCode
+          city
+          householdSize
+          livingSpace
+          norwegianSubType: subType
```
vonElfvin, line 48:
```suggestion
          norwegianHomeContentSubType: subType
```

src/graphql/get-quotes.graphql
```diff
@@ -38,6 +38,20 @@ query GetQuotes($memberId: ID!) {
...
+        }
+
+        ... on NorwegianTravelQuoteData {
+          householdSize
+          isYouth
```
vonElfvin, line 53: If this works that's ok but I don't think we have to do it like this given we have a separate DTO from `NorwegianTravelQuoteData` in back-office, the translation could be there instead. I.e. We could remove `isYouth` and have a subType on `NorwegianTravelQuoteData` just like we do at `NorwegianHomeContentQuoteData`. Let back-office do the translation so to speak.

vonElfvin:
```suggestion
          norwegianTravelSubType: subType
```

src/graphql/use-quotes.tsx
```diff
@@ -44,6 +44,19 @@ export const QUOTES_QUERY = gql`
...
+            city
+            householdSize
+            livingSpace
+            norwegianSubType: subType
+          }
+
+          ... on NorwegianTravelQuoteData {
+            householdSize
+            isYouth
+          }
```
vonElfvin, line 59: Why is there one `get-quotes` and one `use-quotes`?

src/components/member/tabs/quote-tab/quote-modification.tsx
vonElfvin: Instead of ts-ignore

PR #172, opened 2020-05-25
exclude expired quotes

meletios:


src/components/member/tabs/quote-tab/quote-sub-section.tsx
```diff
@@ -20,29 +20,30 @@ export const QuotesSubSection: React.FunctionComponent<{
...
-                })
-              }
-              setIsWip(false)
-            }}
-          />
-        </ActionsWrapper>
-      )}
+      {quotes.filter((quote) => !signedOrExpiredPredicate(quote)).length ===
+        0 && <Button onClick={() => setIsWip(!isWip)}>Create</Button>}
+      {quotes.filter((quote) => !signedOrExpiredPredicate(quote)).length ===
```
palmenhq, line 25: this is getting quite long and this filter method is a few times. Shall we maybe move it to some kind of `signableQuotes` variable or so?

PR #176, opened 2020-05-28
claim transcriptions

FredrikAreschoug:


src/components/claims/claim-details/components/ClaimTranscriptions.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <h3>Transcription</h3>
+    <MuiList>
+      {transcriptions.map((transcription) => (
+        <ListItem>
+          <ClaimTranscription component="p">
+            {transcription.text}
+          </ClaimTranscription>
+          <MuiList>
+            <ClaimTranscriptionMetaData component="span">
+              Confidence: {transcription.confidenceScore}\n Language code:
```
palmenhq, line 53: pretty sure this should be `<br />` not `\n` right?

src/components/claims/claim-details/index.tsx
```diff
@@ -82,6 +84,16 @@ const ClaimPage: React.SFC<Props> = ({ ...props }) => (
                 refetchPage={refetch}
               />
             </Grid>
+            <Grid item xs={12}>
+              {transcriptions && (
+                <ClaimTranscriptions
+                  transcriptions={
+                    (transcriptions.filter(Boolean) as ClaimTranscription[]) ??
```
palmenhq, line 91: i think `(transcriptions?.filter(Boolean) as ClaimTranscription[]) ?? []` would be more appropriate because if the `filter` method already exists we know already that `transcriptions` is not nullish

PR #183, opened 2020-06-09
total number of claims field added

gustafvh:


src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -89,6 +89,23 @@ const MemberInformation: React.SFC<{ member: Member }> = ({ member }) => {
...
+        <span style={{ fontSize: '32px' }}>
+          <FraudulentStatus
+            stateInfo={{
+              state: member.person.debtFlag
+                .replace('GREEN', 'NOT_FRAUD')
+                .replace('AMBER', 'SUSPECTED_FRAUD')
+                .replace('RED', 'CONFIRMED_FRAUD'),
+            }}
+          />
+        </span>
```
vonElfvin, line 107: Hmm, this is not super pretty, the <FraudulentStatus> component should probably be more generic than just taking the fraud status. I do however not think it's that important to fix for now.

schema.graphql
```diff
@@ -58,16 +58,16 @@ type MutationType {
...
 
   addAgreementFromQuote(
-    id: ID!
-    contractId: ID!
-    activeFrom: LocalDate
-    activeTo: LocalDate
+    id: ID!,
+    contractId: ID!,
+    activeFrom: LocalDate,
+    activeTo: LocalDate,
```
vonElfvin, line 70: Was this you who added the commas?

PR #184, opened 2020-06-11
fix monetary amount display for Cost Deduction incentive

AlexSJ1990:


src/lib/intl.ts
```diff
@@ -1,12 +1,16 @@
...
 
-export const formatMoney = (locale: 'sv-SE') => (money: Money) =>
+export const formatMoney = (locale: string) => (money: Money) =>
   new Intl.NumberFormat(locale, {
     style: 'currency',
     currency: money.currency,
-  }).format(typeof money.amount === "number" ? money.amount : parseFloat(money.amount))
+  }).format(
+    typeof money.amount === 'number' ? money.amount : parseFloat(money.amount),
+  )
```
vonElfvin, line 12: Even though this may be objectively correct, to have "kr 10.00" for Norway and "10.00 kr" for Sweden I'm wondering if just having "10.00 NOK" and "10.00 SEK" is more clear. There is another formatMoney in utils/money that does exactly that.

palmenhq: i kinda agree here. Maybe we should kill this `formatMoney` and just use the other `formatMoney` we have inhttps://github.com/HedvigInsurance/hedgehog/blob/master/src/utils/money.tsx

PR #186, opened 2020-06-12
Allow activation of quotes when the contract is pending

vonElfvin:


src/components/member/tabs/quote-tab/quote-activation.tsx
```diff
@@ -66,57 +75,67 @@ export const QuoteActivation: React.FC<{
...
+            <Checkbox
+              onChange={(_, { checked }) => {
                 if (onWipChange) {
                   onWipChange(true)
                 }
-                setPreviousAgreementActiveTo(value)
+                if (!checked) {
+                  setPreviousAgreementActiveTo(null)
+                }
+                setUseGap(checked!)
```
palmenhq, line 106: maybe `Boolean(checked)` instead just to be sure its always a boolean and not null or whatever?

src/components/member/tabs/quote-tab/quote-activation.tsx
```diff
@@ -126,7 +145,7 @@ export const QuoteActivation: React.FC<{
           fullWidth
           disabled={addAgreementMutation.loading}
         >
-          Activate
+          {contract.hasPendingAgreement ? 'Replace' : 'Activate'}
```
palmenhq, line 148: Good ux!

src/components/member/tabs/quote-tab/quote-activation.tsx
```diff
@@ -31,9 +34,14 @@ export const QuoteActivation: React.FC<{
...
-  const [contracts] = useContracts(memberId)
+  const [contracts, { loading }] = useContracts(memberId)
   const contract = getContract(contracts, quote)
-  const [activeFrom, setActiveFrom] = React.useState<Date>(new Date())
+  if (!loading && !contract) {
+    return <>Cannot active quote without Originating product id</>
+  }
+  const [activeFrom, setActiveFrom] = React.useState<Date | null>(
+    getInitialActiveFrom(contract),
+  )
```
palmenhq, line 44: coming back to this we should probably put this inside a getter function instead of passing `new Date` directly to `useState` like so

```ts
useState(() => getInitialActiveFrom(contract))
```

palmenhq: also, while we're poking around in this file maybe we should remove the `React.` prefix from hooks? Because we're using esmoduleinterop now it is no longer needed and we can do `useState(...)` `useEffect(...)` and by importing them like so:

```ts
import React, { useState, useEffect } from 'react'
```

src/components/member/tabs/quote-tab/quote-activation.tsx
vonElfvin: This is a checkbox and I figure null could still behave like false if that were to be the case? I changed `setUseCap(checked!)` to ` setUseCap(!!checked)` to reflect that, what do you think?

src/components/member/tabs/quote-tab/quote-activation.tsx
vonElfvin: Thanks! 😄

src/components/member/tabs/quote-tab/quote-activation.tsx
vonElfvin: Fixed

PR #189, opened 2020-06-17
Fix/fraud status reformat

gustafvh:


shared/hedvig-ui/orb-indicator.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { Icon } from 'semantic-ui-react'
+
+const IconStyled = styled(Icon)(() => ({
+  verticalAlign: 'middle',
+}))
+
+export function OrbIndicator(props) {
+  return (
+    <p style={{ marginTop: '-7px' }}>
+      <b>{props.text}</b>
```
vonElfvin, line 12: Text should probably be outside of the component for the consumer to decide how it should look

shared/hedvig-ui/orb-indicator.stories.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export default {
+  title: 'OrbIndicator',
+  component: OrbIndicator,
+  decorators: [],
+}
+
+export const OrbStatusIndicator: React.FunctionComponent = () => {
+  return <OrbIndicator />
+}
```
vonElfvin, line 12: Maybe include some examples here, try to start storybook to see how this will end up :) "yarn storybook"

shared/hedvig-ui/orb-indicator.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import styled from 'react-emotion'
+import { Icon } from 'semantic-ui-react'
+
+const IconStyled = styled(Icon)(() => ({
+  verticalAlign: 'middle',
+}))
+
+export function OrbIndicator(props) {
```
vonElfvin, line 9: Please use arrow function instead, also specify the props

src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -93,21 +94,11 @@ const MemberInformation: React.SFC<{ member: Member }> = ({ member }) => {
...
-                    .replace('AMBER', 'SUSPECTED_FRAUD')
-                    .replace('RED', 'CONFIRMED_FRAUD'),
-                }}
-              />
-            )}
-          </span>
-        </p>
+        <OrbIndicator
+          text={'Debt Status: '}
+          color={member.person.debtFlag?.toLowerCase()}
```
vonElfvin, line 99: `toLowerCase()` could be moved to the component

shared/hedvig-ui/orb-indicator.tsx
```diff
@@ -0,0 +1,46 @@
...
+interface OrbProps {
+  color?: Flag | string
+  size?: IconSizeProp
+}
+
+enum Flag {
+  GREEN = 'green',
+  AMBER = 'orange',
+  RED = 'red',
+}
```
vonElfvin, line 26: Even though is really neat. I think it's a lot cleaner to just have the regular flag as input and change it to amber inside the component when needed. As I said before having one Generic `OrbIndicator` that takes `SemanticCOLORS` as input and one specific Flag `OrbIndicatorOfFlag` that takes a `Flag` as input which in turn **uses the generic one**  would make this code very easy to read and consume for both the generic cases and  the flag cases. We are not limited to having one component for one use case. Having two component that both show an OrbIndicator is very good for clarity.

shared/hedvig-ui/orb-indicator.tsx
```diff
@@ -0,0 +1,46 @@
...
+  size,
+}) => {
+  return (
+    <Container>
+      <span style={{ fontSize: '32px' }}>
+        <IconStyled
+          name="circle"
+          color={
+            (color?.toLowerCase() as SemanticCOLORS) ||
+            ('grey' as SemanticCOLORS)
```
vonElfvin, line 39:
```suggestion
            (color?.toLowerCase() || 'grey') as SemanticCOLORS)
```

shared/hedvig-ui/orb-indicator.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import styled from 'react-emotion'
+import { Icon } from 'semantic-ui-react'
+import { SemanticCOLORS } from 'semantic-ui-react/dist/commonjs'
```
palmenhq, line 4: pretty sure `import { SemanticCOLORS } from 'semantic-ui-react` should be sufficient?

src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -6,6 +6,7 @@ import * as React from 'react'
 import styled from 'react-emotion'
 import { Link } from 'react-router-dom'
 import { formatMoney } from 'utils/money'
+import { OrbIndicator } from '../../../../../shared/hedvig-ui/orb-indicator'
```
palmenhq, line 9: use `hedvig-ui/orb-indicator` directly instead of all the `..`s :)

src/components/member/tabs/DebtTab.tsx
```diff
@@ -10,6 +10,7 @@ import { RouteComponentProps } from 'react-router'
 import { Table } from 'semantic-ui-react'
 import actions from 'store/actions'
 import { formatMoney } from 'utils/money'
+import { OrbIndicator } from '../../../../shared/hedvig-ui/orb-indicator'
```
palmenhq, line 13: same, no `..`

PR #190, opened 2020-06-18
Feature/graphql questions tab

vonElfvin:
Use graphql instead of api calls via redux to get questions, also fix images/gifs and remove unnecessary code
image


src/components/member/messages/Message.js
```diff
@@ -73,6 +73,14 @@ const Video = styled.video`
   width: 350px;
 `
 
+const Image = styled.img`
+  width: 350px;
```
palmenhq, line 77: `max-width` would probably be more appropriate here :)

src/components/questions/questions-list/Question.tsx
```diff
@@ -0,0 +1,29 @@
...
+import React from 'react'
+import styled from 'react-emotion'
+
+const QuestionWrapper = styled.div<{ isVisible: boolean }>`
+  opacity: ${(props) => (props.isVisible ? 1 : 0)};
+  transform: ${(props) => (props.isVisible ? 'scale(1)' : 'scale(0.8)')};
+  transition: all 0.8s ease-out;
+`
+
+export const Question = ({ question }) => {
```
palmenhq, line 11: i think we should give this some appropriate types: `const Question: React.FC<{ question: QuostionOrSomething }>`

src/components/questions/questions-list/Question.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const [isVisible, setVisible] = React.useState(false)
+
+  React.useEffect(() => {
+    setVisible(true)
+  })
+
+  return (
+    <QuestionWrapper key={question.id} isVisible={isVisible}>
+      <Message
+        content={JSON.parse(question.messageJsonString).body}
```
palmenhq, line 21: should we have this inside some kind of try/catch for example if `messageJsonString` would be empty or unparseable? Just so the whole things doesn't 💥 if so

src/components/questions/questions-list/Question.tsx
```diff
@@ -0,0 +1,29 @@
...
+        left={false}
+        isQuestionMessage={true}
+        timestamp={question.timestamp}
+        from={null}
```
palmenhq, line 25: we should probably make `from` optional so it can be undefined, then we could just leave out `from` completely

src/components/questions/questions-list/QuestionGroupInfo.js
```diff
@@ -0,0 +1,52 @@
...
+const MemberInfoWrapper = styled.div`
+  font-size: 1.5rem;
+  padding-bottom: 1rem;
+`
+
+const QuestionGroupInfo = ({ questionGroup }) => {
+  const memberQuery = useMemberNameAndContractMarketInfoQuery({
+    variables: { memberId: questionGroup.memberId },
+  })
+
```
palmenhq, line 25: wouldn't a better (and better graphql practice) approach be to just query `member` on `questionGroup` directly? so the schema would look something like

```graphqls
type QuestionGroup {
  id: ID!
  member: Member!
  questions: [Question!]!
}
```

and use it like

```ts
questinGroup.member.firstName
```

that would mean we wouldn't have to shoot off so darn many queries  😬

src/containers/questions-page/QuestionsPage.js
```diff
@@ -3,23 +3,11 @@ import { ListPage } from 'components/shared'
...
-})
-
-export default withRouter(
-  connect(mapStateToProps, {
-    ...actions.clientActions,
-    ...actions.questionsActions,
-    ...actions.membersActions,
-  })(QuestionsPage),
-)
+export default withRouter(connect()(QuestionsPage))
```
palmenhq, line 13: we can remove `connect` now

src/store/initialState.ts
```diff
@@ -24,7 +24,7 @@ const initialState: BackofficeStore = {
     member: null,
   },
   members: {
-    list: [],
+    questionGroups: [],
```
palmenhq, line 27: i think we can remove this now no?

vonElfvin: Actually this is the case of some very weird refactoring on intelliJ's part. I have no idea how this happened tbh. Now I understand why the Breadcrumb broke

PR #191, opened 2020-06-22
Minor improvements+fix PR comments

vonElfvin:


src/store/initialState.ts
```diff
@@ -24,7 +24,7 @@ const initialState: BackofficeStore = {
     member: null,
   },
   members: {
-    questionGroups: [],
+    list: [],
```
vonElfvin, line 27: I have no idea how this happened tbh. It should still be there for now I think so we don't break legacy stuff until we want to

palmenhq: Haha! Maybe this is unrelated

PR #193, opened 2020-06-23
Use graphQL instead of socket for the chat

vonElfvin:
Uses GraphQL instead of a socket to send/retrieve messages, also increases the height of the chat. There are some boyscouting to be made in the code itself but I think I've done enough improvements to take it out for a spin to see that it is actually working as intended.
image

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -14,6 +14,7 @@ type QueryType {
   getAnswerSuggestion(question: String): [Suggestion!]!
   me: String
   switchableSwitcherEmails: [SwitchableSwitcherEmail!]!
+  messageHistory(memberId: ID!): [ChatMessage!]!
   questionGroups: [QuestionGroup!]!
   itemCategories(kind: ItemCategoryKind!, parentId: ID): [ItemCategory!]!
   claimItems(claimId: ID!): [ClaimItem!]!
@@ -96,6 +97,7 @@ type MutationType {
   changeToDate(agreementId: ID!, request: ChangeToDateInput): ID!
   changeFromDate(agreementId: ID!, request: ChangeFromDateInput): ID!
   regenerateCertificate(agreementId: ID!): ID!
+  sendMessage(input: SendMessageInput!): SendMessageResponse!
   markQuestionAsResolved(memberId: ID!): Boolean!
   answerQuestion(memberId: ID!, answer: String!): Boolean!
   createQuoteForNewContract(
@@ -1345,6 +1347,28 @@ enum NorwegianGripenFactorType {
   SQUARE_METERS
   HOUSEHOLD_TYPE
   DEDUCTIBLE
+}
+
+type ChatMessage {
+  messageAsJson: String!
+}
+
+union SendMessageResponse = SendMessageSuccessful | SendMessageFailed
+
+type SendMessageSuccessful {
+  memberId: String!
+}
+
+type SendMessageFailed {
+  memberId: String!
+  errorCode: Int!
+  errorMessage: String!
+}
+
+input SendMessageInput {
+  memberId: ID!
+  message: String!
+  forceSendMessage: Boolean!
 }
 
 type MonetaryAmountV2 {
--- a/schema.graphql
+++ b/schema.graphql
@@ -1350,7 +1350,11 @@ enum NorwegianGripenFactorType {
 }
 
 type ChatMessage {
-  messageAsJson: String!
+  globalId: ID!
+  author: String!
+  fromId: String!
+  timestamp: Instant
+  messageBodyJsonString: String!
 }
 
 union SendMessageResponse = SendMessageSuccessful | SendMessageFailed
--- a/src/components/member/chat/ChatPanel.tsx
+++ b/src/components/member/chat/ChatPanel.tsx
@@ -64,7 +64,7 @@ export const ChatPanel = ({ memberId }) => {
   const [forceSendMessage, setForceSendMessage] = useState(false)
   const [sendMessage, { loading }] = useSendMessage()
 
-  const shouldSubmit = (e: React.KeyboardEvent<any>) => {
+  const shouldSubmit = (e: React.KeyboardEvent<HTMLInputElement>) => {
     return (
       !window.matchMedia('(max-width: 800px)').matches &&
       e.keyCode === 13 &&
--- a/src/components/member/chat/ChatPanel.tsx
+++ b/src/components/member/chat/ChatPanel.tsx
@@ -1,12 +1,12 @@
 import {
-  Button as MuiButton,
   FormControlLabel as MuiFormControlLabel,
-  Icon as MuiIcon,
   Switch as MuiSwitch,
   TextField as MuiTextField,
   withStyles,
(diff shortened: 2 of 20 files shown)


src/components/member/chat/ChatPanel.tsx
```diff
@@ -62,119 +59,92 @@ const TextField = withStyles({
...
-            Send
-            <MuiIcon>send</MuiIcon>
-          </SubmitButton>
-        </OptionsContainer>
-      </MessagesPanelContainer>
-    )
-  }
-
-  private shouldSubmit = (e: React.KeyboardEvent<any>) => {
+  const shouldSubmit = (e: React.KeyboardEvent<any>) => {
```
palmenhq, line 67: Shouldn't this be `React.KeyboardEvent<HTMLInputElement>`?

src/components/member/messages/MessagesList.js
```diff
@@ -23,81 +25,61 @@ const getAuthor = (author) => {
...
+  const [messagesList, setMessagesList] = useState(null)
 
-  componentDidMount() {
-    if (this.props.attachScrollListener) {
-      this.props.attachScrollListener(this.theRealScroller)
+  const scrollToBottom = () => {
+    const lastMessage = messages[messages.length - 1]
+    const lastMessageElement = document.getElementById(
+      `msg-${lastMessage.globalId}`,
+    )
```
palmenhq, line 36: ouff, i haven't seen this before. It is very poor practice in react and we should really fix it imo. Rather than querying the document we should use refs.

src/components/member/messages/MessagesList.js
```diff
@@ -23,81 +25,61 @@ const getAuthor = (author) => {
...
+  useEffect(() => {
+    if (messages && messagesList) {
+      scrollToBottom()
     }
-  }
-
-  render() {
-    const { messages, error } = this.props
-    const memberId = parseInt(this.props.id, 10)
+  }, [messages])
```
palmenhq, line 49: what about instead passing for example `messages.length` as a dependency? just `messages` will check object identity, meaning it doesn't care if the array _looks_ the same only if it is the exact same array. And in this particular case we only care about if a new message pops in anyways, right?

src/components/member/messages/MessagesList.js
```diff
@@ -23,81 +25,61 @@ const getAuthor = (author) => {
...
-          <EmptyList>
-            {error
-              ? 'No messages with this Member'
-              : 'Lost connection to server'}
-          </EmptyList>
-        )}
-      </MessagesListContainer>
-    )
+  if (loading) {
+    return null
```
palmenhq, line 52: Unsure if this is what we want. If we refetch now for example the message list will disappear shortly until the refetch is done. Wouldn't a better way be to check if we don't have any content then return null, otherwise return the content?

src/components/member/messages/MessagesList.js
```diff
@@ -23,81 +25,61 @@ const getAuthor = (author) => {
...
-      messagesLength: 0,
-    }
-  }
-
-  componentDidUpdate() {
-    this.theRealScroller()
-  }
+export const MessagesList = ({ memberId }) => {
+  const [messages, {loading}] = useMessageHistory(memberId)
+  const [messagesList, setMessagesList] = useState(null)
```
palmenhq, line 30: you could also just use [`useRef`](https://reactjs.org/docs/hooks-reference.html#useref)

src/graphql/get-message-history.graphql
```diff
@@ -0,0 +1,5 @@
+query GetMessageHistory($memberId: ID!) {
+  messageHistory(memberId: $memberId) {
+    messageAsJson
+  }
+}
```
palmenhq, line 5: i think we might get some caching issues with this, since it doesn't have an id. If we _really_ cant avoid the json-in-json Is there any chance we could get IDs in here, or could we somehow extract it from the `messageAsJson` and pass it to [`dataIdFromObject`](https://github.com/HedvigInsurance/hedgehog/blob/master/src/api/apollo-client.ts#L29-L36)?

src/graphql/use-message-history.tsx
```diff
@@ -0,0 +1,24 @@
+import {
+  GetMessageHistoryQueryHookResult,
+  useGetMessageHistoryQuery,
+} from 'api/generated/graphql'
+
+type MessageHistoryReturnTuple = [
+  object[] | undefined,
```
palmenhq, line 7: the `object` literal means an object  with no properties. Is that really what we mean here? Wouldn't `unknown[]` be more appropriate?

schema.graphql
```diff
@@ -1347,6 +1349,28 @@ enum NorwegianGripenFactorType {
   DEDUCTIBLE
 }
 
+type ChatMessage {
+  messageAsJson: String!
```
palmenhq, line 1353: Is there no way we can avoid json-in-json? I feel very reluctant to this type of api design, as it gets virtually impossible to understand or validate what will be in here and requires extra work on the client. Unless there's a very good reason this is kind of a dealbreaker for me.

src/graphql/use-message-history.tsx
```diff
@@ -0,0 +1,24 @@
...
+]
+
+export const useMessageHistory = (
+  memberId: string,
+): MessageHistoryReturnTuple => {
+  const queryResult = useGetMessageHistoryQuery({
+    variables: {
+      memberId,
+    },
+    pollInterval: 2000,
```
sampettersson, line 18: Just an FYI last time we tried this in the app bot service started to crash as some users chats are quite heavy to load, so can recommend to test it thoroughly with large chats before releasing hehe 😄

src/components/member/chat/ChatPanel.tsx
vonElfvin: Yes I can fix that

src/components/member/messages/MessagesList.js
vonElfvin: That did not work but I changed some other stuff so it works now :) Changed it to messages?.length

src/components/member/messages/MessagesList.js
vonElfvin: It doesn't dissappear on refetch actually. I could add both though, to check `if (loading && !messages)` to be sure

src/graphql/use-message-history.tsx
vonElfvin: Hmm, do we have any other approach so solve this?

src/graphql/use-message-history.tsx
```diff
@@ -0,0 +1,34 @@
...
+  })
+  const messageHistory = queryResult?.data?.messageHistory as
+    | Message[]
+    | undefined
+
+  messageHistory?.forEach(
+    (message) =>
+      ((message as Message).body = JSON.parse(message.messageBodyJsonString)),
+  )
+  return [messageHistory, queryResult]
```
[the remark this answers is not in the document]
vonElfvin: Awesome, I tried doing this but completely forgot about the spread operator

PR #194, opened 2020-06-25
Use GraphlQL to get dashboard numbers to be able to kill all sockets

vonElfvin:


src/components/questions/questions-list/QuestionGroup.tsx
```diff
@@ -14,6 +14,7 @@ const QuestionGroupWrapper = styled.div<{ isVisible: boolean }>`
   border-radius: 0.5rem;
   max-width: 50rem;
   width: 50%;
+  min-width: 350px;
```
vonElfvin, line 17: An attempt to fix the iPhone issue. I'll test it and see if it's an upgrade or not

PR #196, opened 2020-06-30
Add automated depreciation and valuation

rasmusguterstam:
**Automated depreciation** 📱

- Gives suggested valuation (with explanation) using a specified depreciation table
- Possible to use different depreciation tables depending on type of contract

**Valuation applicable**
usage_1

**No explicit valuation available**
usage_2


src/components/claims/claim-details/components/claim-items/components/ItemForm.tsx
```diff
@@ -26,40 +32,59 @@ export const ItemForm: React.FC<{
...
     itemFamilyId: selectedItemCategories[0]?.id ?? null,
     itemTypeId: selectedItemCategories[1]?.id ?? null,
     itemBrandId: selectedItemCategories[2]?.id ?? null,
     itemModelId: selectedItemCategories[3]?.id ?? null,
-    dateOfPurchase: dateOfPurchase === '' ? null : dateOfPurchase,
-    purchasePriceAmount: purchasePrice !== '' ? Number(purchasePrice) : null,
-    purchasePriceCurrency: purchasePrice !== '' ? purchasePriceCurrency : null,
+    dateOfPurchase:
+      dateOfPurchase !== '' && dateOfPurchase.length === 10
+        ? dateOfPurchase
```
vonElfvin, line 53: I think we should send it down even though it is not 10 characters to crash it, the UX itself should make it hard but otherwise we will just 🙈 when they have incorrect dates and send nothing instead. Rather crash IMO.

src/components/claims/claim-details/components/claim-items/components/ItemForm.tsx
```diff
@@ -7,13 +7,19 @@ import { format, isAfter, isValid, parseISO } from 'date-fns'
...
 
 const isValidDate = (date: string) =>
   date === ''
     ? true
     : isValid(parseISO(date)) && !isAfter(parseISO(date), new Date())
 
+const isNotValidDate = (date: string) => !isValidDate(date)
+const isValidNumber = (n: string) => /^[0-9]*$/g.test(n)
+const isNotValidNumber = (n: string) => !isValidNumber(n)
+const isNotEmpty = (s: string | null) => s !== '' && s !== null
```
vonElfvin, line 20: I personally think !isValidDate is explicit enough making "isNotValidDate" a unnecessary function to have. Similar case to the others.

src/components/claims/claim-details/components/claim-items/components/ItemForm.tsx
```diff
@@ -26,40 +32,59 @@ export const ItemForm: React.FC<{
...
+        : null,
+    purchasePriceAmount: isNotEmpty(purchasePrice)
+      ? Number(purchasePrice)
+      : null,
+    purchasePriceCurrency: isNotEmpty(purchasePrice)
+      ? purchasePriceCurrency
+      : null,
+    valuationAmount: Number(autoValuation) ?? null,
+    customValuationAmount:
+      isNotEmpty(customValuation) && isValidNumber(customValuation)
```
vonElfvin, line 63: Similarly, the form itself should still send the `customValuation` down even though it is invalid for the backend to break. Otherwise we send down nothing even though they made a type for example `1000b` we send down `null` which is weird

vonElfvin: The form itself should validate it of course, maybe disable the button and make the correct input red or something. But the send to the backend should not ignore the error.

src/components/claims/claim-details/components/claim-items/components/ItemForm.tsx
```diff
@@ -26,40 +32,59 @@ export const ItemForm: React.FC<{
...
+        ? Number(customValuation)
+        : null,
+    valuationCurrency:
```
vonElfvin, line 66: Thinking of this, maybe we should ask IEX if they sometimes evaluate in different currencies than the one it was purchased in. For example if you buy a phone in France for €400 and we evaluate it to be worth 2000 SEK?

vonElfvin: That would have to be a custom evaluation of course

src/components/claims/claim-details/components/claim-items/components/ItemForm.tsx
```diff
@@ -68,70 +93,93 @@ export const ItemForm: React.FC<{
...
+              style: {
+                paddingLeft: '10px',
+              },
+            }}
           />
         </Grid>
       </Grid>
       <Grid container spacing={16}>
-        <Grid item style={{ width: '79.5%' }} />
+        <Grid item style={{ width: '79.5%' }}>
```
vonElfvin, line 173: 79.5% is very close to 80% which could be used in a xs={}

src/components/claims/claim-details/components/claim-items/components/ItemForm.tsx
```diff
@@ -26,40 +32,59 @@ export const ItemForm: React.FC<{
...
-  const validPurchasePrice = /^[0-9]*$/g.test(purchasePrice)
   const defaultCurrency = contractMarketInfo?.preferredCurrency ?? 'SEK'
 
   const [
     upsertClaimItem,
     { loading: loadingUpsertClaimItem },
   ] = useUpsertClaimItemMutation()
 
-  const addNewClaimItemRequest: UpsertClaimItemInput = {
+  const formData: UpsertClaimItemInput = {
```
vonElfvin, line 45: Since this is used for the request only I would name it after the variable, either:
```suggestion
  const request: UpsertClaimItemInput = {
```
Or
```suggestion
  const input: UpsertClaimItemInput = {
```

src/components/claims/claim-details/components/claim-items/components/MessageChip.tsx
```diff
@@ -0,0 +1,149 @@
...
+  ] = useCanEvaluate('SE_APARTMENT_RENT', itemFamilyId, itemTypeId)
+
+  const [evaluation, { loading: loadingEvaluation }] = useGetEvaluation(
+    price ?? 0,
+    itemFamilyId,
+    'SE_APARTMENT_RENT',
+    dateOfPurchase,
+    itemTypeId,
+    null,
+  )
```
vonElfvin, line 52: I think this should be performed outside the `MessageChip` and the evaluation itself will be input to this component

src/components/claims/claim-details/components/claim-items/components/MessageChip.tsx
```diff
@@ -0,0 +1,149 @@
...
+      return <NoValuationChip />
+    }
+  }
+
+  return (
+    <>
+      {getCurrentChip()}
+      {itemFamilyId && (
+        <>
+          <InputChip
```
vonElfvin, line 132: This one should probably be its own component that is also used by the parent of MessageChip. The two chips speak to eachother only through the inputs they get from the parent.

src/components/claims/claim-details/components/claim-items/components/MessageChip.tsx
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    if (canEvaluate) {
+      return <InfoChip />
+    }
+
+    if (itemFamilyId && !loadingEvaluationStatus) {
+      return <NoValuationChip />
+    }
+  }
```
vonElfvin, line 125: This is more or less everything the MessageChip needs to do

src/components/claims/claim-details/components/claim-items/components/MessageChip.tsx
vonElfvin: A parent "EvaluationChips" that coordinates all of this makes sense to me. With one "MessageChip" and one "CustomEvaluationChip"

src/components/claims/claim-details/components/claim-items/components/chips/components/DiscardChip.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+
+import styled from 'react-emotion'
+import { BaseChip } from './BaseChip'
+
+const Chip = styled(BaseChip)`
```
vonElfvin, line 6: Consider if this can be extracted to `hedvig-ui` for other usages going forward

src/components/claims/claim-details/components/claim-items/components/ItemForm.tsx
vonElfvin: I.e. the automatic evaluation does not need to care about other currencies

src/components/claims/claim-details/components/claim-items/components/MessageChip.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  React.useEffect(() => {
+    setCustomValuation('')
+  }, [itemFamilyId, itemTypeId])
+
+  const getExplanation = (
+    evaluationRule: EvaluationRule | null | undefined,
+  ) => {
+    if (typeof evaluationRule === 'undefined' || evaluationRule === null) {
```
palmenhq, line 77: i think we should either use `typeof`, or not - we shouldn't mix them :) the reason for using `typeof` is mainly if you're not sure the variable exists in the current scope yet, which is not really the case here since we know we'll have it from 2 lines above. So `evaluationRule === undefined || evaluationRule === null` should be sufficent here. Really good to check for both though!

src/components/claims/claim-details/components/claim-items/components/MessageChip.tsx
```diff
@@ -0,0 +1,149 @@
...
+      return null
+    }
+
+    const evaluationName = evaluationRule?.evaluationName
```
palmenhq, line 81: I don't think the `?.` operator here is needed because typescript should figure out that it is there due to the nullish check above

palmenhq: and i guess the same applies to all `?.` operators here really

src/components/claims/claim-details/components/claim-items/components/chips/components/BaseChip.tsx
```diff
@@ -0,0 +1,42 @@
...
+`
+
+export const BaseChip: React.FC<{
+  label: string
+  icon?: React.ReactElement | undefined
+  adornment?: React.ReactElement | undefined
+  clickable?: boolean
+  className?: string
+  onDelete?: React.EventHandler<any>
+  onClick?: React.EventHandler<any>
```
palmenhq, line 20: Maybe we could get these from `ChipProps` instead? Even though they're the same we should probably use the source directly. like so

```ts
import { ChipProps} from '@material-ui/core/Chip'

// ...
export const BaseChip: React.FC<{ 
  // ...
  onDelete?: ChipProps["onDelete"]
```

palmenhq: actually, should we not just pass all the chip props? So one can use whatever from the material ui props.. You can forward all props with something like this:

```tsx
 import { Chip } from '@material-ui/core' 
 import { ChipProps } from '@material-ui/core/core/Chip' 

export const BaseChip: React.FC<ChipProps> = (props) => {
  return (
    <Chip
      {...props}
      avatar={icon ? <ChipIconBase>{icon}</ChipIconBase> : undefined} 
    />
  )
}
```
or something in the lines of this. Might need some more customization but you get the point

src/components/claims/claim-details/components/claim-items/components/chips/components/DiscardChip.tsx
```diff
@@ -0,0 +1,22 @@
...
+  background-color: rgba(0, 0, 0, 0);
```
palmenhq, line 7: you could also use `transparent` here, might make it more obvious what the intention is

src/components/claims/claim-details/components/claim-items/components/chips/components/InfoChip.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { LightningFill } from 'react-bootstrap-icons'
+import styled from 'react-emotion'
+import { BaseChip } from './BaseChip'
+
+const Chip = styled(BaseChip)`
+  background: ${({ theme }) => theme.success};
+  color: white;
```
palmenhq, line 8: this should probably be `theme.accentContrast` no?

src/components/claims/claim-details/components/claim-items/components/chips/components/InputChip.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import { Pencil } from 'react-bootstrap-icons'
+import styled from 'react-emotion'
+
+import { TextField } from '@material-ui/core'
+import { BaseChip } from './BaseChip'
+
+const Chip = styled(BaseChip)`
+  background-color: ${({ theme }) => theme.accent};
+  color: white;
```
palmenhq, line 10: `accentContrast`

src/components/claims/claim-details/components/claim-items/components/chips/components/InputChip.tsx
```diff
@@ -0,0 +1,64 @@
...
+  margin-top: 8px;
+  font-weight: bold;
+  margin-left: 7px;
+  width: 130px;
+  cursor: pointer;
+`
+
+const textFieldStyle = {
+  border: '0px',
+  backgroundColor: 'rgba(0,0,0,0)',
```
palmenhq, line 20: i think this should be `transparent` again.

src/components/claims/claim-details/components/claim-items/components/chips/components/MarketValuationChip.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { Gem } from 'react-bootstrap-icons'
+import styled from 'react-emotion'
+import { BaseChip } from './BaseChip'
+
+const Chip = styled(BaseChip)`
+  background-color: ${({ theme }) => theme.success};
+  color: white;
```
palmenhq, line 8: you know the drill

src/components/claims/claim-details/components/claim-items/components/chips/components/DiscardChip.tsx
```diff
@@ -0,0 +1,22 @@
...
+  border: 1px solid ${({ theme }) => theme.accent};
+`
+
+const DiscardChip: React.FC<{ onClick: React.EventHandler<any> }> = ({
+  onClick,
+}) => {
+  return <Chip label={'Reset'} onClick={onClick} />
+}
+
+export default DiscardChip
```
palmenhq, line 22: we try to avoid `export default` and rather use named exports, like `export const DiscardChip`, instead. This is for several reasons;
 - named exports are - well - named, meaning it's way easier to find their usage and makes intellisense in your editor confused (for example you could import this like `import AReallyBadName from './DIscardChip'`)
- default exports interop poorly with [CommonJS](https://en.wikipedia.org/wiki/CommonJS), a module system we're not using for this project but might need to use elsewhere and we should keep our code standards as neat as possible
 - default exports interop poorly with [code splitting](https://webpack.js.org/guides/code-splitting/) and async imports. Conside:

```ts
import('./DiscardChip').then(module => module.default).then(DiscardChip => { /* do something */ })
// vs
import('./DiscardChip').then(({ DiscardChip }) => { /* do something */ })
```

[here's](https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/) some additional reading on the subject if you're interested :)

src/components/claims/claim-details/components/claim-items/components/chips/index.tsx
```diff
@@ -0,0 +1,6 @@
+export { default as DiscardChip } from './components/DiscardChip'
```
palmenhq, line 1: you're already experiencing the pains of default exports here i can see :) `export * from './components/DiscardChip'`  would probably be sufficient without them

src/components/claims/claim-details/components/claim-items/components/styles.tsx
```diff
@@ -1,6 +1,7 @@
-import { Chip, Select } from '@material-ui/core'
+import { Chip } from '@material-ui/core'
 import styled from 'react-emotion'
 import CreatableSelect from 'react-select/creatable'
+import { Popover } from '../../../../../../../shared/hedvig-ui/popover'
```
palmenhq, line 4: no need for `../../.././......../shared` here, `hedvig-ui/popover` is enough

src/components/claims/claim-details/components/claim-items/components/MessageChip.tsx
```diff
@@ -0,0 +1,149 @@
...
+      ? '...'
+      : evaluation?.depreciatedValue?.toString() ?? '-',
+    currency: currency ?? defaultCurrency,
+  }
+
+  React.useEffect(() => {
+    setCustomValuation('')
+  }, [itemFamilyId, itemTypeId])
+
+  const getExplanation = (
```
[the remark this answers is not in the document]
vonElfvin: We have a utils folder for things like this, you can maybe add a file there for ClaimItem

src/components/claims/claim-details/components/claim-items/components/chips/components/BaseChip.tsx
rasmusguterstam: Really good point!

src/components/claims/claim-details/components/claim-items/components/ItemForm.tsx
rasmusguterstam: I see what you mean, but that's not really what happens in practice. Instead, this makes sure that getClaimItemValuation isn't fired until it actually has a proper date (i.e. it makes sure that "2019" gets converted to "2019-01-01" by using the fuzzy date feature. The first case would lead to getClaimItemValuation using "2019" as date input, which obv. is invalid)

src/components/claims/claim-details/components/claim-items/components/ItemList.tsx
```diff
@@ -43,35 +42,21 @@ export const ItemList: React.FC<{ claimId: string }> = ({ claimId }) => {
...
       <colgroup>
-        <col style={{ width: '50.7%' }} />
-        <col style={{ width: '16.6%' }} />
-        <col style={{ width: '13.4%' }} />
-        <col style={{ width: '16.3%' }} />
-        <col style={{ width: '3.0%' }} />
+        <col style={{ width: '50.5%' }} />
+        <col style={{ width: '17.0%' }} />
+        <col style={{ width: '9.0%' }} />
+        <col style={{ width: '20.0%' }} />
```
palmenhq, line 52: How come the total width is not just 100%?

src/components/claims/claim-details/components/claim-items/components/ValuationInfo.tsx
```diff
@@ -0,0 +1,68 @@
...
+  setCustomValuationAmount: React.EventHandler<any>
+  defaultCurrency: string
+}> = ({
+  request,
+  setValuation,
+  customValuationAmount,
+  setCustomValuationAmount,
+  defaultCurrency,
+}) => {
+  const TYPE_OF_CONTRACT = 'SE_APARTMENT_RENT' // To be changed
```
palmenhq, line 21: Why? When? By who?

src/components/claims/claim-details/components/claim-items/components/chips/MessageChip.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { getValuationExplanation } from 'utils/claim-item'
+import { ExplanationPopover } from '../styles'
+import { InfoChip } from './components/InfoChip'
+import { MarketValuationChip } from './components/MarketValuationChip'
+import { NoValuationChip } from './components/NoValuationChip'
+import { ValuationChip } from './components/ValuationChip'
+
+export const MessageChip: React.FC<{
+  valuation: ClaimItemValuation | undefined
+  valuationStatus: CanValuateClaimItem | undefined
```
palmenhq, line 16: You don't need to fix it here, but to save a few keystrokes in the future you could also write this as `valuation?: ClaimItemValuation`

src/components/claims/claim-details/components/claim-items/components/chips/MessageChip.tsx
```diff
@@ -0,0 +1,67 @@
...
+  dateOfPurchase,
+  customValuation,
+}) => {
+  const loadingValuation = valuation === undefined
+  const priceAndDateAvailable = price && dateOfPurchase?.length === 10
+  const canValuateClaimItem = !!itemFamilyId && !!valuationStatus?.canValuate
+  const valuationType = valuation?.valuationRule?.valuationType ?? ''
+  const marketValuation = valuationType === 'MARKET_PRICE'
+  const formattedValuation: MonetaryAmountV2 = {
+    amount: valuation?.depreciatedValue?.amount.toString() ?? '...',
```
palmenhq, line 37: should `amount` really be `...`...?

src/components/claims/claim-details/components/claim-items/components/chips/components/BaseChip.tsx
```diff
@@ -0,0 +1,39 @@
...
+  onDelete,
+  onClick,
+  clickable,
+  className,
+}) => {
+  return (
+    <Chip
+      label={label}
+      avatar={icon ? <ChipIconBase>{icon}</ChipIconBase> : undefined}
+      onDelete={onDelete ?? (adornment && (() => void 0))}
```
palmenhq, line 32: I haven't heard of the `void` operator before - really clever use of it! NICE!

src/components/claims/claim-details/components/claim-items/components/ValuationInfo.tsx
```diff
@@ -11,29 +11,31 @@ export const ValuationInfo: React.FC<{
...
   const { itemFamilyId, itemTypeId, purchasePrice, dateOfPurchase } = request
 
-  const [valuationStatus] = useCanValuateClaimItem(
-    itemFamilyId,
-    itemTypeId,
-    null,
-  )
+  const valuationStatus =
+    typeOfContract &&
+    useCanValuateClaimItem(itemFamilyId, itemTypeId, typeOfContract)[0]
```
palmenhq, line 27: you can't do this unfornately. Hooks can't be conditional, as per [react docs](https://reactjs.org/docs/hooks-overview.html#rules-of-hooks):

> Hooks are JavaScript functions, but they impose two additional rules:
>    - Only call Hooks at the top level. Don’t call Hooks inside loops, conditions, or nested functions.

it will result in 💥  eventually

src/components/claims/claim-details/components/claim-items/components/chips/index.tsx
```diff
@@ -0,0 +1,6 @@
+export * from './components/DiscardChip'
+export * from './components/InfoChip'
+export * from './components/InputChip'
+export * from './components/MarketValuationChip'
+export * from './components/NoValuationChip'
+export * from './components/ValuationChip'
```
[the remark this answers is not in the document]
rasmusguterstam: It _was_ used to enable multiple imports from ./chip, but things changed and you're right - it's not necessary anymore!

src/components/claims/claim-details/components/claim-items/components/ValuationInfo.tsx
```diff
@@ -0,0 +1,68 @@
...
+    typeOfContract,
+  )
+
+  const [claimItemValuation] = useGetClaimItemValuation({
+    purchasePrice: purchasePrice ?? 0,
+    itemFamilyId: itemFamilyId ?? null,
+    typeOfContract,
+    purchaseDate: dateOfPurchase,
+    itemTypeId: itemTypeId ?? null,
+    baseDate: null,
```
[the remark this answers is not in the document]
rasmusguterstam: I get your point, but if you implement the check inside the helper function (like the sample code below) you get an issue with changed orders of hooks 😕

```
if (!request.purchasePrice || !request.itemFamilyId) {
    return [undefined, undefined]
  }

const queryResult = useGetClaimItemValuationQuery({ ...
```

PR #207, opened 2020-07-14
Kill old questions, messages, use graphQL instead of messages to pass around member

vonElfvin:
Dependent on https://github.com/HedvigInsurance/back-office/pull/282


src/utils/member.tsx
```diff
@@ -42,11 +42,7 @@ export const getMemberIdColor = (memberId: string) => {
   return lightTheme.success
 }
 
-export const getMemberGroup = (memberId: number) => {
-  if (memberId === tomasMemberId) {
-    return 'The dark lord Sith'
-  }
```
vonElfvin, line 48: :(

src/components/claims/claim-details/components/ClaimInformation.tsx
```diff
@@ -208,12 +189,15 @@ const ClaimInformation: React.SFC<Props> = ({
             <MuiMenuItem disabled value={'none'} divider>
               None selected
             </MuiMenuItem>
-            {contracts.map((_contract) => (
-              <MuiMenuItem key={_contract.id} value={_contract.id}>
-                {_contract.contractTypeName} (
-                {getAddressFromContract(_contract)})
-              </MuiMenuItem>
-            ))}
+            {contracts.map((_contract) => {
```
palmenhq, line 192: if we're poking around here we might as well remove the leading underscore. It's normally used to name an unused variable but this one is used. If it's because there is something defined called `contract` in the scope already that's usually denoted with _trailing_ underscore

src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -56,9 +62,12 @@ const MemberInformation: React.SFC<{ member: Member }> = ({ member }) => {
       <p>
         <b>Personal Number:</b> {member.personalNumber}
       </p>
-      <p>
-        <b>Address:</b> {member.address}, {member.postalNumber} {member.city}
-      </p>
+      {address && (
+        <p>
+          <b>Address:</b> {address.street}, {address.postalCode} {address.city}
```
palmenhq, line 67: `strong`, never ever `b`

src/components/claims/claim-details/index.tsx
```diff
@@ -1,7 +1,7 @@
 import Grid from '@material-ui/core/Grid'
 import { ClaimNote, ClaimTranscription, QueryType } from 'api/generated/graphql'
 import { ClaimItems } from 'components/claims/claim-details/components/claim-items'
-import ChatPane from 'components/member/tabs/ChatPane.js'
+import { ChatPane } from 'components/member/tabs/ChatPane.js'
```
palmenhq, line 4: no need for `.js` suffix

src/components/member/index.js
```diff
@@ -64,126 +67,99 @@ const MemberDetail = styled.span`
...
 const MemberDetailLink = MemberDetail.withComponent('a')
 
-export default class Member extends React.Component {
-  getMemberPageTitle = (member) =>
-    `${member && (member.firstName || '') + ' ' + (member.lastName || '')}`
+export const Member = (props) => {
+  const memberId = props.match.params.memberId
+  const [member, { loading }] = useGetMemberInfo(memberId)
+  const getMemberPageTitle = (member) =>
+    `${(member.firstName || '') + ' ' + (member.lastName || '')}`
```
palmenhq, line 74: why are we using both a template string  and `' ' + ' '`? I think we should pick either

src/components/member/index.js
```diff
@@ -64,126 +67,99 @@ const MemberDetail = styled.span`
...
-      claimsByMember,
-    } = this.props
+  useEffect(() => {
+    props.claimsByMember(memberId)
+  }, [])
 
-    memberRequest(memberId)
-    claimsByMember(memberId)
-  }
+  if (loading) return null
```
palmenhq, line 80: brackets plz :)

src/components/member/index.js
```diff
@@ -64,126 +67,99 @@ const MemberDetail = styled.span`
...
-    const {
-      match: {
-        params: { memberId },
-      },
-      memberRequest,
-      claimsByMember,
-    } = this.props
+  useEffect(() => {
+    props.claimsByMember(memberId)
+  }, [])
```
palmenhq, line 78: just an idea should we put `memberId` as a dependency for `useEffect`? That way this request will fire every time the member id changes, otherwise if the member id would change the claims would remain the same i guess.

src/components/member/index.js
```diff
@@ -64,126 +67,99 @@ const MemberDetail = styled.span`
...
-                    >
-                      {messages?.member?.memberId}
-                    </MemberDetailLink>
-                  </Popover>
-                </MemberDetails>
+  const panes = memberPagePanes(props, memberId, member)
+  return (
+    <MemberHistoryContext.Consumer>
+      {({ pushToMemberHistory }) => (
+        <Mount on={() => pushToMemberHistory(memberId)}>
```
palmenhq, line 86: imo we should do this in an `useEffect` now that we have access to it instead

src/components/member/index.js
```diff
@@ -64,126 +67,99 @@ const MemberDetail = styled.span`
...
-                            format: 'text/plain',
-                          },
-                        )
-                      }}
-                    >
-                      {messages?.member?.memberId}
-                    </MemberDetailLink>
-                  </Popover>
-                </MemberDetails>
+  const panes = memberPagePanes(props, memberId, member)
```
palmenhq, line 82: should we inline this variable? Its only used once so its a bit confusing to create it up here imo

src/components/member/index.js
```diff
@@ -64,126 +67,99 @@ const MemberDetail = styled.span`
...
+      )}
+    </MemberHistoryContext.Consumer>
+  )
 }
 
 Member.propTypes = {
-  messageReceived: PropTypes.func.isRequired,
   match: PropTypes.object.isRequired,
-  messages: PropTypes.object.isRequired,
   auth: PropTypes.object.isRequired,
```
palmenhq, line 162: unused no?

src/components/member/index.js
palmenhq: Also we could probably move this function out somewhere else since its  pure, to slim this component a bit

src/components/member/index.js
```diff
@@ -64,126 +67,99 @@ const MemberDetail = styled.span`
 `
 const MemberDetailLink = MemberDetail.withComponent('a')
 
-export default class Member extends React.Component {
-  getMemberPageTitle = (member) =>
-    `${member && (member.firstName || '') + ' ' + (member.lastName || '')}`
+export const Member = (props) => {
```
palmenhq, line 70: What about destructing the props here so its easier to see which props are expected? (so this'd be something like `({ match: { params: { memberId } } }) => {`)

src/components/member/tabs/ChatPane.js
```diff
@@ -35,65 +35,40 @@ const ChatHeaderStyle = styled.div`
...
 
-  resizeControlChat = (e) => {
-    if (!this.state.manualChange) {
-      this.setState({ visible: window.innerWidth > 1500 })
+  const resizeControlChat = (e) => {
+    if (!manualChange) {
+      setVisible(window.innerWidth > 1500)
     }
   }
+  window.addEventListener('resize', resizeControlChat)
```
palmenhq, line 47: this should be done within a `useEffect` so it can be reversed once this component unmounts or we'll have explosions on other pages

```js
useEffect(() => {
  const resizeControlChat = (e) => {
    if (!manualChange) {
       setVisible(window.innerWidth > 1500) 
    }
  }
  window.addEventListener('resize', resizeControlChat)

  return () => window.removeEventListener('resize', resizeControlChat) // reverse when component unmounts
}, [])
```

src/components/member/tabs/index.js
```diff
@@ -40,16 +40,19 @@ TabItem.propTypes = {
   hideTab: PropTypes.bool,
 }
 
-const memberPagePanes = (props) => {
-  const memberId = props.match.params.memberId
+const memberPagePanes = (props, memberId, member) => {
```
palmenhq, line 43: do we use these props for anything now? I think we shouldn't pass the props around like this unless we _really_ really need to

src/components/member/index.js
vonElfvin: I think we'll be able to do that very soon, right now we send the whole "props" to the panes below which I'll change soon but not now

src/components/member/tabs/index.js
vonElfvin: We rarely use it, but we still use it on some pages. It'll have to be in a future refactoring push though.

PR #208, opened 2020-07-15
Clean/remove dashboard

vonElfvin:
Note: Should merge the Questions/Messages first so this one is not so large :)


src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -56,9 +62,13 @@ const MemberInformation: React.SFC<{ member: Member }> = ({ member }) => {
...
       </p>
-      <p>
-        <b>Address:</b> {member.address}, {member.postalNumber} {member.city}
-      </p>
+      {address && (
+        <p>
+          <b>Address:</b> {address?.street}, {address?.postalCode}{' '}
+          {address.city}
+        </p>
+      )}
```
rasmusguterstam:
```suggestion
      {address && (
        <p>
          <b>Address:</b> {address.street}, {address.postalCode}{' '}
          {address?.city}
        </p>
      )}
```
Since the schema is `type Address {
  street: String!
  postalCode: String!
  city: String
}` and we're doing null check with `address &&` 🤪

src/components/claims/claim-details/index.tsx
```diff
@@ -76,7 +76,12 @@ const ClaimPage: React.SFC<Props> = ({ ...props }) => (
                   )}
 
                   <Grid item xs={12} sm={12} md={4}>
-                    {member && <MemberInformation member={member} />}
+                    {member && (
+                      <MemberInformation
+                        member={member}
+                        contract={contract ?? undefined}
+                      />
+                    )}
```
rasmusguterstam, line 84: Are we defaulting to `undefined` when contract is `undefined` here? Or is it if contract is `null`? In either case, seems a bit off to explicitly default to `undefined` 🧐

src/components/member/tabs/ChatPane.js
```diff
@@ -35,65 +35,38 @@ const ChatHeaderStyle = styled.div`
...
+        <ChatHeader visible={visible} onResizeClick={onResizeClick} />
+        <MessagesList memberId={memberId} />
+        <ChatPanel memberId={memberId} />
+      </Resizable>
+    </>
+  ) : (
+    <>
+      <ChatHeader visible={visible} onResizeClick={onResizeClick} />
+    </>
+  )
```
rasmusguterstam:
```suggestion
  return visible ? (
      <Resizable
        style={resizableStyles}
        defaultSize={{ width: '400px', height: '80%' }}
        enable={{ left: true }}
      >
        <ChatHeader visible={visible} onResizeClick={onResizeClick} />
        <MessagesList memberId={memberId} />
        <ChatPanel memberId={memberId} />
      </Resizable>
  ) : <ChatHeader visible={visible} onResizeClick={onResizeClick} />
```💅

File: src/components/member/tabs/ChatPane.js [hunk 4]
```diff
@@ -109,5 +82,5 @@ const ChatHeader = (props) => (
 )
 
 ChatPane.propTypes = {
-  match: PropTypes.object.isRequired,
+  memberId: PropTypes.string.isRequired,
 }
```
rasmusguterstam, line 88: Why are we doing this, isn't this redundant? Can we not define the entire component as: 
```
export const ChatPane: React.FunctionComponent<{
  memberId: string
}> = ({ memberId }) => {
```

rasmusguterstam: I know it _was_ a class component, but figured why not since we're refactoring to a functional one!

src/components/member/tabs/index.js
```diff
@@ -86,18 +89,12 @@ const memberPagePanes = (props) => {
   panes.push(
```
rasmusguterstam, line 89: We should just skip this and define the entire list in the beginning!

src/utils/member.tsx
```diff
@@ -42,11 +42,7 @@ export const getMemberIdColor = (memberId: string) => {
   return lightTheme.success
 }
 
-export const getMemberGroup = (memberId: number) => {
-  if (memberId === tomasMemberId) {
-    return 'The dark lord Sith'
-  }
```
rasmusguterstam, line 48: 😢

src/components/member/tabs/ChatPane.js
vonElfvin: Mostly becuase it is javascript I think 🤷

PR #210, opened 2020-07-17
Fixing & Cleaning: Member debt tab

rasmusguterstam:
### What is this
Noticed a weird overflow for the debt tab, and had some time to kill. Ended up being a makeover with overall design as well. And making it functional. And using hooks. And autogenerated graphql. And some other stuff. 💅

Mostly code and design fixes though, it is still using the same logic!

### Before
Screenshot 2020-07-17 at 10 05 10

### After
Screenshot 2020-07-17 at 10 04 59

### After, with button
image


src/components/member/tabs/debt-tab/WhitelistMemberButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const [whitelistMember] = useWhitelistMemberMutation()
+
+  return (
+    <Button
+      variation="primary"
+      fullWidth
+      onClick={() => {
+        const confirm = window.confirm('Are you sure?')
+
+        if (confirm) {
```
palmenhq, line 19: `if (!confirm) { return }`?

src/components/member/tabs/debt-tab/WhitelistMemberButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+  showNotification: (data: Notification) => void
+}> = ({ memberId, showNotification }) => {
+  const [whitelistMember] = useWhitelistMemberMutation()
+
+  return (
+    <Button
+      variation="primary"
+      fullWidth
+      onClick={() => {
+        const confirm = window.confirm('Are you sure?')
```
palmenhq, line 17: Maybe a more verbose message? Something like `are you sure you want to whitelist "Blargis"?`

src/graphql/get-person.graphql
```diff
@@ -0,0 +1,33 @@
+query GetPerson($memberId: ID!) {
+  member(id: $memberId) {
+    memberId
+    person {
+      debtFlag
```
palmenhq, line 5: do we have any IDs or personnummer or something we can use to let the apollo cache understand which person is which?

src/components/member/tabs/debt-tab/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+                  : 'Not Whitelisted'}
+              </InfoText>
+            </InfoRow>
+          </InfoContainer>
+          <Spacing top={'large'} />
+          {eligibleForWhitelist && (
+            <WhitelistMemberButton
+              memberId={memberId}
+              showNotification={showNotification}
+            />
```
palmenhq, line 127: Can i suggest we wrap `WhitelistMemberButtton` with `WithNotification` instead of passing it like this? :)

src/components/member/tabs/index.js
```diff
@@ -13,6 +12,7 @@ import { Tab } from 'semantic-ui-react'
 import styled from 'react-emotion'
 import { useContractMarketInfo } from 'graphql/use-get-member-contract-market-info'
 import { AccountTab } from './account-tab'
+import { DebtTab } from 'components/member/tabs/debt-tab'
```
palmenhq, line 15: maybe we should import this like `./deb-tab` instead?

src/graphql/get-person.graphql
palmenhq: Optionally if we configure the cache to normalize the person in some other way. See https://www.apollographql.com/docs/react/caching/cache-configuration/

rasmusguterstam: I looked into this a bit, and if I'm not completely off the `person` type is always used under the type `member`, such that it's cached along with member (and in turn `memberId`, which we have specified in 'apollo-client.ts'). You never get the `person` without going through `member`. 

With that said, I believe the naming here is pretty off since the `person` type only contains debt-related information, but I don't feel like changing every implementation currently using it 😅

src/components/member/tabs/shared/card-components.tsx
```diff
@@ -0,0 +1,20 @@
+import { Paragraph } from 'hedvig-ui/typography'
+import styled from 'react-emotion'
+
```
AlexSJ1990, line 3: nice 😍

PR #216, opened 2020-08-03
Feature/question group country sort

gustafvh:
- [x]  Modify Resolver for QuestionGroup, so that it contains member, which in turn contains Contract Market info.
- [x]  Add country-flag on each question group
- [x]  Fix sorting-function on frontend to sort on market
- [x]  Member contains first and last name meaning refactor so it takes it from there instead

Branch was very behind master so a large marge was done with merge conflict in schema.graphql

Fix was needed in backend and is already pushed: https://github.com/HedvigInsurance/back-office/pull/290/files


src/components/questions/questions-list/QuestionGroupInfo.js
```diff
@@ -19,18 +19,15 @@ const MemberInfoWrapper = styled.div`
...
   return (
     <QuestionGroupInfoWrapper memberId={questionGroup.memberId}>
       <MemberInfoWrapper>
         {memberDataMaybe ? (
           <>
             {memberDataMaybe?.firstName} {memberDataMaybe?.lastName}{' '}
-            {getMemberFlag(memberDataMaybe?.contractMarketInfo?.market)},{' '}
+            {getMemberFlag(memberDataMaybe?.contractMarketInfo?.market)}
           </>
         ) : null}
```
rasmusguterstam, line 32: Maybe I'm missing something and there's a good reason to why, but could we do some better naming here? 🧐
And also skip `: null` and instead go for something like:

```suggestion
  const member = questionGroup?.member
  return (
    <QuestionGroupInfoWrapper memberId={questionGroup.memberId}>
      <MemberInfoWrapper>
        {member && 
          <>
            {member.firstName} {member.lastName}{' '}
            {getMemberFlag(member.contractMarketInfo?.market)},{' '}
            {getMemberFlag(member.contractMarketInfo?.market)}
          </>
          }
```

src/components/questions/questions-list/QuestionGroups.js
```diff
@@ -12,10 +12,14 @@ const ListContainer = styled.div`
...
 `
 
-const doFilter = (selectedFilters) => ({ memberId }) =>
+const doTeamFilter = (selectedFilters) => ({ memberId }) =>
   (selectedFilters.includes(FilterState.Even) && isMemberIdEven(memberId)) ||
   (selectedFilters.includes(FilterState.Odd) && !isMemberIdEven(memberId))
 
+const doMarketFilter = (selectedFilters) => ({ member }) =>
+  (selectedFilters.includes(FilterState.Sweden) && (member.contractMarketInfo?.market === "SWEDEN")) ||
+  (selectedFilters.includes(FilterState.Norway) && (member.contractMarketInfo?.market === "NORWAY"))
```
rasmusguterstam, line 21:
```suggestion
  (selectedFilters.includes(FilterState.Sweden) && (member.contractMarketInfo?.market === Market.Sweden)) ||
  (selectedFilters.includes(FilterState.Norway) && (member.contractMarketInfo?.market === Market.Norway))
```

src/components/questions/questions-list/QuestionGroupInfo.js
gustafvh: I agree, && is better suited here. Nice spot!

PR #221, opened 2020-08-06
Make "Visible No Discount"-campaigns creatable

rasmusguterstam:
## Included
- Implements the option to create "Visible No Discount"-campaigns under campaign tool

image


src/features/tools/campaign-codes/forms/VisibleNoDiscountForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { withShowNotification } from 'utils/notifications'
+
+const initialFormData: AssignVoucherVisibleNoDiscount = {
+  code: '',
+  partnerId: '',
+  validFrom: null,
+  validUntil: null,
+}
+
+const formLooksGood = (formData: AssignVoucherVisibleNoDiscount) => {
```
palmenhq, line 25: maybe we could rename this to `formIsValid`? It's more conventional

src/features/tools/campaign-codes/forms/VisibleNoDiscountForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+  return formData.partnerId !== '' && formData.code !== ''
+}
+
+const VisibleNoDiscount: React.FC<{} & WithShowNotification> = ({
```
palmenhq, line 29: no need for `{} & ...`React.FC<WithShowNotification>` should be sufficient :)

src/features/tools/campaign-codes/forms/VisibleNoDiscountForm.tsx
```diff
@@ -0,0 +1,130 @@
+import { AssignVoucherVisibleNoDiscount } from 'api/generated/graphql'
+import { ClearableDropdown as Dropdown } from 'features/tools/campaign-codes/components/ClearableDropdown'
+import { Centered, Row } from 'features/tools/campaign-codes/styles'
+import { mapCampaignOwners } from 'features/tools/campaign-codes/utils'
```
palmenhq, line 4: Could we use `./(styles|utils)` as import instead? I think that would make it clearer it's related to this package

src/features/tools/campaign-codes/forms/VisibleNoDiscountForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+        disabled={loading}
+        onChange={({ currentTarget: { value: code } }) =>
+          setFormData({ ...formData, code })
+        }
+        placeholder="Code"
+      />
+      <Spacing top={'small'} />
+      <label>Valid period</label>
+      <Row>
+        <div style={{ float: 'left' }}>
```
palmenhq, line 71: I think we should use flexbox for this, floats are sooo 2010 ;)

PR #223, opened 2020-08-13
Cleaning: Make stuff reusable (hedvig-ui)

rasmusguterstam:
## Included

Adds some utilities/components that I've been reusing over and over again, which I believe would be useful in the future as well.

* **`<Badge>`**: A Badge component, simliar to button but even rounder corners and non-clickable
* **`<Capitalized>`**: Make any text within _Capitalized_. Very useful when you have ALL CAPS ENUMS.
* **`<Placeholder>`**: Often there's a scenario where a value is missing or not available. Instead of leaving it empty or using same styling as if there was a value, use a Placeholder component to apply `theme.placeholderColor`: `<Placeholder>Not available</Placeholder>`

For more specifics, check out corresponding story in hedvig-ui!


shared/hedvig-ui/badge.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+const variants: ReadonlyArray<NonNullable<BadgeProps['variant']>> = [
+  'danger',
+  'warning',
+  'success',
+  'default',
+]
+
+export const StandardBadge: React.FC = () => {
+  return (
+    <>
```
vonElfvin, line 21:
```suggestion

```

vonElfvin (dismissed):
Beaut 😍

palmenhq:
But really really nice  😍

vonElfvin (approved):
:)

PR #229, opened 2020-08-26
Use correct id for navigation to claim from questions

vonElfvin:


palmenhq (approved):

src/components/questions/questions-list/QuestionGroupInfo.js
```diff
@@ -48,7 +48,7 @@ const QuestionGroupInfo = ({ questionGroup }) => {
                   disabled={member.claims.length > 1}
                   onClick={() =>
                     history.push(
-                      `/claims/${member.claims[0].id}/members/${member.id}`,
+                      `/claims/${member.claims[0].id}/members/${member.memberId}`,
```
palmenhq, line 51: i would argue this should be a `Link` element though

PR #230, opened 2020-09-01
Add count of number that apply to filter in Questions tab

vonElfvin:
image

### DIFF
--- a/src/components/questions/filter.tsx
+++ b/src/components/questions/filter.tsx
@@ -74,21 +74,17 @@ export const QuestionsFilter: React.FC<{
   const getCountByFilter = (filter: FilterState) => {
     switch (filter) {
       case FilterState.Even:
-        return questionGroups.filter(doTeamFilter([FilterState.Even])).length
+        return questionGroups.filter(doTeamFilter([filter])).length
       case FilterState.Odd:
-        return questionGroups.filter(doTeamFilter([FilterState.Odd])).length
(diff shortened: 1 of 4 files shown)


src/components/questions/filter.tsx
```diff
@@ -61,15 +67,33 @@ const GreenTeamBadge = styled(TeamBadge)`
...
+        return questionGroups.filter(doTeamFilter([filter])).length
+      case FilterState.Sweden:
+        return questionGroups.filter(doMarketFilter([filter])).length
+      case FilterState.Norway:
+        return questionGroups.filter(doMarketFilter([filter])).length
+      case FilterState.HasOpenClaim:
+        return questionGroups.filter(doClaimFilter([filter])).length
+      case FilterState.NoOpenClaim:
+        return questionGroups.filter(doClaimFilter([filter])).length
+    }
```
palmenhq, line 88: Instead of mapping it with switch you could just pass the filter function in no? so the function is used something like

```ts
getCountByFilter(doTeamFilter, FilterState.Odd)
```

and the signature would look something like

```ts
const getCountByFilter =(filterer: (active: FilterState) => boolean, filter: FilterState) => {
  return questionGroups.filter(filterer([filter])).length
}
```
untested but you get the point :)

palmenhq:
Awesome! (sorry, i will turn off circle ci)

palmenhq (approved):
Some beautiful functional programming going on here 😍

PR #234, opened 2020-09-23
Feature: Kill MembersStore and use GraphQL

rasmusguterstam:
### What's up 💡

1. Kills off MembersStore from redux
2. Implements searchMember query (Wooho, finally able to resolve whatever we want 🎉)
3. Implements setFraudulentStatus mutation
4. Changes which data is available when searching for members. 
4.1 Removes unclear concept of "contract status" and uses contract counter instead for each category (Active, Pending or Terminated). 
4.2 Adds flag next to member name
4.3 Removes columns "Active from" and "Active to" (not clear what these imply - what if we have multiple contracts in different states?)

### Images

image

image


src/components/members-search/components/ContractCountCircles.tsx
```diff
@@ -0,0 +1,90 @@
...
+      case ContractStatus.Terminated:
+        acc[status] = (acc[status] || 0) + 1
+        break
+      default:
+        acc[ContractStatus.Active] = (acc[ContractStatus.Active] || 0) + 1
+    }
+
+    return acc
+  }, {})
+
```
palmenhq, line 43: Rather than modifying `acc`, return a new immutable one :) like so:

```ts
  contracts.reduce((acc, { status }) => {
    const groupedStatus = [ContractStatus.Pending, ContractStatus.Terminated].includes(status)
      ? status
      : ContractStatus.Active
    return {
       ...acc,
       [groupedStatus]: (acc[groupedStatus] || 0) + 1
    }
  }, {})
```

src/components/members-search/components/ContractCountCircles.tsx
```diff
@@ -0,0 +1,90 @@
...
+  )
+}
+
+export const ContractCountCircles: React.FC<{
+  contracts: Contract[]
+}> = ({ contracts }) => {
+  const {
+    ACTIVE: ActiveContracts = 0,
+    PENDING: PendingContracts = 0,
+    TERMINATED: TerminatedContracts = 0,
```
palmenhq, line 68: i think we should use camelCase rather than PascalCase for variables, or it may be confused with a type or a component. So this would become something like `activeContracts` or so

src/containers/member-search-page/index.tsx
```diff
@@ -1,27 +1,9 @@
...
-  searchLoading,
-}) => (
-  <MembersSearch
-    searchMemberRequest={searchMemberRequest}
-    searchResult={searchResult}
-    searchLoading={searchLoading}
-  />
-)
+// TODO: Inject new MemberSearchPageComponent here
+// TODO: Shouldn't we generate Member type from graphql:gen instead?
```
palmenhq, line 5: huh?

palmenhq (commented):
really awesome work!

src/containers/member-search-page/index.tsx
rasmusguterstam: Haha oops, old comment 🤦‍♂️

src/components/members-search/components/ContractCountCircles.tsx
rasmusguterstam: That's really clean! 😀

palmenhq (dismissed):
Beautiful 😍

src/utils/contract.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+})
+
+it('getLastTerminationDate returns null first contract is active', () => {
+  const oneActiveOneTerminated = [
+    {
+      isTerminated: false,
+    },
+    {
+      terminationDate: '2020-03-01',
+      isTerminated: true,
```
palmenhq, line 101: aren't these dates `Date`s in the `Contract` type? If so i think we should treat them as such in the tests also or the tests might be very off. And if they really are strings i think the way we compare the dates might be a bit shaky and rather we should parse the date strings into `Date`s :)

palmenhq (commented):
Awesome with some tests :D

src/components/members-search/components/ContractCountCircles.tsx
```diff
@@ -0,0 +1,97 @@
...
+    ].includes(status)
+      ? status
+      : ContractStatus.Active
+    return {
+      ...acc,
+      [groupedStatus]: (acc[groupedStatus] || 0) + 1,
+    }
+  }, {})
+
+export const CountCircle: React.FC<{
```
vonElfvin, line 45: Maybe keep this circle thing in hedvig-ui? :)

src/components/members-search/components/ContractCountCircles.tsx
```diff
@@ -0,0 +1,97 @@
...
+  justify-content: center;
+  align-items: center;
+`
+
+type NumberOfContracts = {
+  [key in ContractStatus]?: number
+}
+
+const countContractsByStatus = (contracts: Contract[]): NumberOfContracts =>
+  contracts.reduce((acc, { status }) => {
```
vonElfvin, line 31: Really neat way to do this 👏

src/components/members-search/components/ListHeader.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { Table } from 'semantic-ui-react'
+
+export const ListHeader: React.FC = () => (
+  <Table.Header>
+    <Table.HeaderCell>Member</Table.HeaderCell>
+    <Table.HeaderCell />
```
vonElfvin, line 7: What is in this one?

src/components/members-search/components/SearchForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+  onSubmit,
+  onFocus,
+  loading,
+  query,
+  setQuery,
+  includeAll,
+  setIncludeAll,
+  currentResultSize,
+  searchFieldRef,
+}) => {
```
vonElfvin, line 34: I think "onFocus" was lost in the merge with master here? Cannot see it being used.

src/graphql/member-search.graphql
```diff
@@ -0,0 +1,22 @@
...
+      contractMarketInfo {
+        market
+      }
+      contracts {
+        status
+        masterInception
+        terminationDate
+      }
+    }
+    page
```
vonElfvin, line 19: Have you tested the pagination?

src/utils/contract.tsx
```diff
@@ -64,3 +64,41 @@ export const getContractByAgreementId = (
...
+    .filter((contract) => !!contract.masterInception)
+    .map((contract) => contract.masterInception)
+  if (masterInceptions.length === 0) {
+    return null
+  }
+  return masterInceptions.reduce((a, b) => {
+    if (a < b) {
+      return a
+    }
+    return b
```
vonElfvin, line 81: I know I wrote this one but 😄 
```suggestion
    return a < b ? a : b
```

vonElfvin (commented):
Good job! Added some things :D

src/components/members-search/components/ListHeader.tsx
rasmusguterstam: It provides an extra column for member info!

image

src/graphql/member-search.graphql
rasmusguterstam: Do you mean as actually writing a test for it? Or just making sure it works?

vonElfvin (approved):
🔥

PR #238, opened 2020-09-30
Add ability to update and create quotes via json schema

vonElfvin:
Update house example:
image

Create apartment example:
image


shared/hedvig-ui/json-schema-form.tsx
```diff
@@ -0,0 +1,208 @@
...
+  schema: object
+  onSubmit: (formData) => void
+  initialFormData?: object
+  submitText?: string
+}> = ({ schema, onSubmit, initialFormData, submitText, children }) => {
+  const uiSchema = {
+    'ui:ObjectFieldTemplate': ObjectFieldTemplate,
+  }
+  const [formData, setFormData] = useState(initialFormData ?? null)
+  console.log(formData)
```
vonElfvin, line 185: No console log

shared/hedvig-ui/json-schema-form.tsx
```diff
@@ -0,0 +1,208 @@
...
+    return error
+  })
+}
+
+const getPropertyTitle = (property) => {
+  return camelcaseToTitleCase(property.substring(1))
+}
+
+const camelcaseToTitleCase = (text) =>
+  text.charAt(0).toUpperCase() + text.substring(1).replace(/(\B[A-Z])/g, ' $1')
```
vonElfvin, line 169: Move to util

src/components/member/tabs/index.js
```diff
@@ -3,6 +3,7 @@ import { DetailsTab } from 'components/member/tabs/DetailsTab'
 import MemberFile from 'components/member/tabs/FileTab'
 import PaymentsTab from 'components/member/tabs/payments-tab'
 import { Quotes } from 'components/member/tabs/quote-tab'
+import { Quotes as QuotesV2 } from 'components/member/tabs/quote-tab-v2'
```
vonElfvin, line 6: Remove

src/components/member/tabs/quote-tab/index.tsx
```diff
@@ -1,6 +1,7 @@
-import { Quote } from 'api/generated/graphql'
+import { ContractType, Quote } from 'api/generated/graphql'
 import { useContractMarketInfo } from 'graphql/use-get-member-contract-market-info'
-import { useQuotes } from 'graphql/use-quotes'
```
vonElfvin, line 3: Clean up this one

src/graphql/get-quotes.graphql
```diff
@@ -0,0 +1,62 @@
...
+      createdAt
+      breachedUnderwritingGuidelines
+      originatingProductId
+      signedProductId
+      isReadyToSign
+      schema
+      schemaData
+      data {
+        ... on ApartmentQuoteData {
+          street
```
vonElfvin, line 22: This should be able to be removed now

vonElfvin: I decided not to remove it for other usages. I'll clean this up all the way when we see that this works.

src/utils/quote.tsx
```diff
@@ -41,3 +42,17 @@ export const getSubType = (quoteData: QuoteData): string => {
...
 }
+
+export const signedOrExpiredPredicate = (quote: Quote) =>
+  expiredPredicate(quote) || signedPredicate(quote)
+
+export const expiredPredicate = (quote: Quote) => {
+  const createdAt = new Date(quote.createdAt)
+  const now = new Date()
+
+  createdAt.setSeconds(quote.validity!)
```
palmenhq, line 53: I see three issues here.

1. I'm pretty sure `setSeconds` i supposed to set the second of the minute, meaning it should be between 0-59. An easier way out i think would be to use `quote.createdAt + quote.validity` and pass that to the constructor, right?

2. I believe the `createdAt` and `validity` timestamps are in seconds right? JavaScript expects all timestamps to be in milliseconds, so assuming the timestamps are in seconds this won't work. You need to multiply them both with `1000` (i'm slightly unsure about this one but i'm kinda sure)

3. What if `validity` is null? We're disregarding that case now.

src/utils/quote.tsx
```diff
@@ -41,3 +42,17 @@ export const getSubType = (quoteData: QuoteData): string => {
   }
   throw Error(`Unable to return subType of quoteData=${quoteData}`)
 }
+
+export const signedOrExpiredPredicate = (quote: Quote) =>
+  expiredPredicate(quote) || signedPredicate(quote)
+
+export const expiredPredicate = (quote: Quote) => {
```
palmenhq, line 49: Please challenge this, i love the word predicate. But i think its mostly used within typing (in the context of "put a predicate here") and i think for these functions could rather be named `isSignedOrExpired` and `isExpired`. Both work though :)

palmenhq (changes requested):
I think you may be lookin for this package :) https://www.npmjs.com/package/@types/json-schema

src/utils/quote.tsx
vonElfvin: I moved this from another file, haven't really considered it tbh. I'll check it out though

src/utils/quote.tsx
vonElfvin: Same here, I just moved this. Can look into it though

src/utils/quote.tsx
vonElfvin: But yea, that line looks off to me in its entirety. I'll just remove it tbh.

src/utils/quote.tsx
vonElfvin: I agree with you, isSigned etc. is way more clear. Didn't really understand what it did with the word "predicate"

vonElfvin: However, I'm pretty sure it's another @palmenhq criticises younger @palmenhq here 😉

src/graphql/use-update-quote-by-schema.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const useUpdateQuoteBySchema = () => useUpdateQuoteBySchemaMutation()
+
+export const getUpdateQuoteSchemaOptions = (
+  memberId: string,
+  quoteId: string,
+  schema: object,
+  formData: object,
+  bypassUnderwritingGuidelines: boolean,
+): MutationFunctionOptions<
```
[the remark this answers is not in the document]
vonElfvin: Didn't know you could do that, been trying multiple times but never seem to figure out where to put the `: Params` declaration! Really neat :)

shared/hedvig-ui/json-schema-form.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+const StyledForm = styled(Form)`
+  width: 100%;
+`
+
+export const JsonSchemaForm: React.FC<{
+  schema: object
+  onSubmit: (formData) => void
+  initialFormData?: object
```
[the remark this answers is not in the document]
vonElfvin: unknown won't work at a later stage, but Record is what I'm looking for here :)

src/components/member/tabs/quote-tab/update-quote-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+const UpdateQuoteFormComponent: React.FC<{
+  quote: Quote
+  onSubmitted: () => void
+} & WithShowNotification> = ({ quote, onSubmitted, showNotification }) => {
+  const [bypassUwgl, setBypassUwgl] = useState(false)
+  const [updateQuote] = useUpdateQuoteBySchema()
+
+  const performQuoteUpdate = (formData) => {
+    const options = getUpdateQuoteSchemaOptions(
+      quote.memberId!!,
```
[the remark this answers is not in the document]
vonElfvin: Kotlin I guess 🤷

src/graphql/use-get-schema-for-contract-type.tsx
```diff
@@ -0,0 +1,22 @@
+import {
+  ContractType,
+  GetSchemaForContractTypeQueryHookResult,
+  useGetSchemaForContractTypeQuery,
+} from 'api/generated/graphql'
+
+type GetSchemaForContractTypeReturnTuple = [
+  object,
```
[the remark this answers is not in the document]
vonElfvin: used `JSONSchema7` :)

src/components/member/tabs/contracts-tab/agreement/CreateQuoteFromAgreement.tsx
```diff
@@ -3,28 +3,30 @@ import {
...
       <ThirdLevelHeadline>Create Quote</ThirdLevelHeadline>
       <>
         {quotes
           .filter(
-            (quote) =>
-              quote.state === QuoteState.Quoted && !expiredPredicate(quote),
+            (quote) => quote.state === QuoteState.Quoted && !isExpired(quote),
           )
           .map((quote) => quote.originatingProductId)
           .includes(agreement.id) && !loadingQuotes ? (
```
rasmusguterstam, line 32: Perhaps this could be defined as a const earlier on, e.g. `const quoteAlreadyExists = ...` to make it more readable? 🙃

src/graphql/use-create-quote-for-member-by-schema.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const getCreateQuoteForMemberBySchemaOptions = ({
+  memberId,
+  schema,
+  formData,
+  bypassUnderwritingGuidelines,
+}: GetCreateQuoteForMemberBySchemaOptionsParams): MutationFunctionOptions<
+  CreateQuoteForMemberBySchemaMutation,
+  CreateQuoteForMemberBySchemaMutationVariables
+> => {
```
rasmusguterstam, line 28: I've been thinking about this approach as to how we write query-helpers. Might just be me, but using a variable helper (`get...Options`) always feels a bit awkward - why require this intermediate step? Why not use lazy queries and integrate it? 

I tried a variant on this [here](https://github.com/HedvigInsurance/hedgehog/pull/234/files?file-filters%5B%5D=.js&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-248b7e23976c26724393abc9b4e64eb2), could this be a valid approach or am I missing something obvious? 🧐

vonElfvin: The reason being that the options for the mutation are usually not super simple to build. For example the one you sent, knowing what defaults are reasonable etc. is not always obvious for the consumer of the mutation. 

For this case in particular the 

```
schemaData: {
   id: schema.$id,
   ...formData,
```
will be a bit hard to figure out as a consumer unless it is centralised like this.

Also having relevant refetchQueries centralised makes sense as well, like for quotes for the particular member in this case.

shared/hedvig-ui/json-schema-form.tsx
```diff
@@ -0,0 +1,205 @@
...
+        break
+      case 'maximum':
+        error.message = `Should be less than ${!error.params.exlusive &&
+          '(or equal to) '}${error.params.limit}`
+        break
+    }
+    error.stack = `${getPropertyTitle(error.property)}: ${error.message}`
+    return error
+  })
+}
```
palmenhq, line 166: Idk why my comment disappeared but i really dislike this way of mutating the existing object. Rather, we should make a copy of it and return that new error as i suggested before.

src/components/member/tabs/contracts-tab/agreement/CreateQuoteFromAgreement.tsx
vonElfvin: Makes sense, I'll move it :)

palmenhq (commented):
Amaze! Just one detail i think is worth fixing

shared/hedvig-ui/json-schema-form.tsx
vonElfvin: Oh yea, I missed that part 🤦 Fixed it now!

palmenhq (dismissed):

PR #243, opened 2020-10-12
Clean up duplicate libraries

palmenhq:
I was gonna do something completely different and really tiny but my 6 core i7 MBP couldn't handle hedgehog anymore so i dug down the rabbit hole of figuring out why. Turns out I was able to shave of a good **44 MEGABYTES(!!!!!!!!!)** of Javascript from the dev build (so the bundle size is now less than half). Yes, you read that right and yes, i double checked the numbers. As a reference, the *full* size of the market web dev build is 10 megabytes

The build times (cached) improved also:
- Dev: from 1 minute 7 seconds -> 32 seconds
- Prod: a massive 2 mins 27 seconds  -> 1 min 1 second.

And - my macbook doesn't freeze up when i try to develop hedgehog now.

(a tip, check this PR without whitespace diff) 

# Bundle stats (compressed)

## Dev (after)

Screenshot 2020-10-12 at 14 28 47

## Dev (before)
Screenshot 2020-10-12 at 14 31 58

## Prod (after)
Screenshot 2020-10-12 at 14 39 57

## Prod (before)
Screenshot 2020-10-12 at 14 35 44


src/lib/helpers.ts
```diff
@@ -189,9 +188,7 @@ function sortListByDate(list, fieldName, isReverse) {
   })
 
   const sortedDates = withDates.sort((a, b) => {
-    const dateA = moment(a[fieldName] || '10000-01-01')
-    const dateB = moment(b[fieldName] || '10000-01-01')
-    return dateA.diff(dateB)
+    return a - b
```
rasmusguterstam, line 191: Haha w o w 🌟

rasmusguterstam (approved):

src/lib/helpers.ts
palmenhq: Gotta save those bytes

.babelrc.js
```diff
@@ -1,24 +1,28 @@
...
-  ],
-  env: {
-    test: {
-      plugins: [
-        '@babel/transform-modules-commonjs',
-        '@babel/plugin-transform-runtime',
-        'babel-plugin-dynamic-import-node',
-      ],
+module.exports = (api) => {
+  api.cache(true)
```
vonElfvin, line 2: A stupid question. What will this do? 😄

src/App.js
```diff
@@ -95,85 +93,76 @@ class App extends React.Component {
...
+                        <Routes.PrivateRoute
+                          path="/tools"
+                          store={store}
+                          component={Routes.ToolsPageRoute}
+                        />
+                        <Routes.PrivateRoute
+                          path="/taskmanager"
+                          store={store}
+                          component={Routes.TaskManagerPageRoute}
+                        />
```
vonElfvin, line 147: If you wanna kill some more bytes we can kill this whole thing all the way through 🔪

src/App.js
```diff
@@ -95,85 +93,76 @@ class App extends React.Component {
...
+                        <Route
+                          path="/ticket_history/:id"
+                          render={(routeProps) => (
+                            <Routes.TicketHistoryPageRoute
+                              {...routeProps}
+                              store={store}
+                            />
+                          )}
+                        />
```
vonElfvin, line 156: This as well of course

package.json
```diff
@@ -46,8 +45,6 @@
     "koa-server-http-proxy": "^0.1.0",
     "koa-static": "^5.0.0",
     "matchly": "^1.0.1",
-    "material-ui-pickers": "^2.0.1",
-    "moment": "^2.22.2",
```
vonElfvin, line 50: Finally 😍

src/routes/routes.ts
```diff
@@ -14,6 +14,6 @@ export const reactPageRoutes = [
     Component: Routes.TicketHistoryPageRoute,
     exact: true,
   },
-  { path: '/dashboard', Component: Routes.DashboardPageRoute, exact: true },
+  { path: '/dashborad', Component: Routes.DashboardPageRoute, exact: true },
```
vonElfvin, line 17: Haha 😄

src/server/auth.ts
```diff
@@ -35,7 +35,7 @@ export const loginCallback: Middleware<object> = async (ctx) => {
     method: 'POST',
   })
 
-  ctx.redirect('/dashboard')
+  ctx.redirect('/dashborad')
```
vonElfvin, line 38: I can predict someone will refresh their page at dashborad and get pranked once this update is out 😄

vonElfvin (approved):
Boom 💥 💣

.babelrc.js
palmenhq: I'm slightly unsure exactly what it does but i guess it kicks in some kind of cache mechanism in babel and ultimately speeds up the sequential builds 👍

rasmusguterstam (approved):

PR #248, opened 2020-10-20
Consistent loading messages, smoother transitions

rasmusguterstam:
### Loading and Error feedback
If you navigate H.ope for a couple of minutes, you will notice that every change of screen, query and mutation handles loading and error feedback differently. For example, you will find awkward messages saying: "Loading", "Loading...", "loading", "Something went wrong" etc., often as an unformatted string. This is without doubt convenient, but it's not pretty nor remarkable. To handle this, with inspiration from @palmenhq's lovely member search animations, we now have `MajorMessage` and `MajorLoadingMessage` (simply a `MajorMessage` with animated ellipsis at the end). 

Let's just not do `if (loading) { return <div>Loading...</div> }` anymore. 
Do `if (loading) { return <MajorLoadingMessage>Loading</MajorLoadingMessage> }`

### Smooth transitions between screens
On the same note, we have a lot of different screens / routes where content should be updated once data has been retrieved. Currently, the data pops up whenever it's available, but I feel we could do better (and as already mentioned, why not strive for the same 'wow'-feeling you get at `/dashborad` and `/members` 🌟). With that said, animations could escalate. So I kept it simple and introduced a wrapper component, `EaseIn`, to simply get smooth transitions for whatever you want. 

##### Disclaimer
Honestly don't know if these animations would hurt performance or if I'm missing something obvious. Just a suggestion, I think it looks pretty neat, but perhaps it is _too much_ if we have it everywhere 🤓 @palmenhq 

##### To do before merge ✅
- Add Storyblok examples

### DIFF
--- a/shared/hedvig-ui/animations/ease-in.tsx
+++ b/shared/hedvig-ui/animations/ease-in.tsx
@@ -1,0 +1,13 @@
+import styled, { keyframes } from 'react-emotion'
+
+const fadeIn = (max) =>
+  keyframes({
+    from: { opacity: 0, transform: 'translateY(2%)' },
+    to: { opacity: max, transform: 'translateY(0)' },
+  })
+
+export const EaseIn = styled('div')(() => ({
+  opacity: 0,
+  animation: `${fadeIn(1.0)} 1000ms forwards`,
+  animationDelay: '0ms',
+}))
--- a/shared/hedvig-ui/animations/ease-in.tsx
+++ b/shared/hedvig-ui/animations/ease-in.tsx
@@ -6,8 +6,10 @@ const fadeIn = (max) =>
     to: { opacity: max, transform: 'translateY(0)' },
   })
 
-export const EaseIn = styled('div')(() => ({
-  opacity: 0,
-  animation: `${fadeIn(1.0)} 1000ms forwards`,
-  animationDelay: '0ms',
-}))
+export const EaseIn = styled('div')<{ delay?: string }>(
+  ({ delay = '0ms' }) => ({
+    opacity: 0,
+    animation: `${fadeIn(1.0)} 1000ms forwards`,
+    animationDelay: delay,
+  }),
+)
--- a/shared/hedvig-ui/animations/major-message.tsx
+++ b/shared/hedvig-ui/animations/major-message.tsx
@@ -1,0 +1,81 @@
+import React from 'react'
+import styled, { keyframes } from 'react-emotion'
+
+const fadeIn = (max) =>
+  keyframes({
+    from: { opacity: 0, transform: 'translateY(2%)' },
+    to: { opacity: max, transform: 'translateY(0)' },
+  })
+
+const MajorMessageWrapper = styled('div')<{
+  paddingTop?: string
+  paddingBottom?: string
+  paddingLeft?: string
+  paddingRight?: string
+}>(({ paddingTop = '25vh', paddingBottom, paddingLeft, paddingRight }) => ({
+  display: 'flex',
+  flexDirection: 'column',
+  justifyContent: 'center',
+  alignItems: 'center',
+  textAlign: 'center',
+  opacity: 0,
+  animation: `${fadeIn(0.3)} 1000ms forwards`,
+  animationDelay: '20ms',
+  width: '100%',
+  flex: 1,
+  fontSize: '1.5rem',
+  paddingTop,
+  paddingBottom,
+  paddingLeft,
+  paddingRight,
+}))
+
+const AnimatedEllipsis = styled.span`
+  &::after {
+    display: inline-block;
+    animation: ellipsis 1.5s infinite;
+    content: '.';
+    width: 1em;
+    text-align: left;
+  }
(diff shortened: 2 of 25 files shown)


shared/hedvig-ui/animations/ease-in.tsx
```diff
@@ -0,0 +1,15 @@
+import styled, { keyframes } from 'react-emotion'
+
+const fadeIn = (max) =>
```
vonElfvin, line 3: fadeIn or easeIn 🤔

vonElfvin:
```suggestion
const fadeIn = (max: number) =>
```

shared/hedvig-ui/animations/major-message.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react'
+import styled, { keyframes } from 'react-emotion'
+
+const fadeIn = (max) =>
+  keyframes({
+    from: { opacity: 0, transform: 'translateY(2%)' },
+    to: { opacity: max, transform: 'translateY(0)' },
+  })
```
vonElfvin, line 8: Maybe move this to some animation file to be used by both "ease-in" and "major-message"

shared/hedvig-ui/animations/ease-in.tsx
```diff
@@ -0,0 +1,15 @@
...
+  keyframes({
+    from: { opacity: 0, transform: 'translateY(2%)' },
+    to: { opacity: max, transform: 'translateY(0)' },
+  })
+
+export const EaseIn = styled('div')<{ delay?: string }>(
```
vonElfvin, line 9: I think 🤔 we should use the `` syntax for styled components. @palmenhq would have to correct me on that though.
```suggestion
export const EaseIn = styled('div')<{ delay?: string }>`
```

shared/hedvig-ui/animations/major-message.tsx
```diff
@@ -0,0 +1,91 @@
...
+      paddingBottom={paddingBottom}
+      paddingLeft={paddingLeft}
+      paddingRight={paddingRight}
+    >
+      {children}
+    </MajorMessageWrapper>
+  )
+}
+
+export const MajorLoadingMessage: React.FC<MajorMessageProps> = ({
```
vonElfvin, line 81: Having a "Major" loading message speaks to me like we have a "Minor" as well. Maybe it could just be called `LoadingMessage`?

shared/hedvig-ui/animations/major-message.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+interface MajorMessageProps {
+  children: React.ReactNode
+  paddingTop?: string
+  paddingBottom?: string
+  paddingLeft?: string
+  paddingRight?: string
+}
+
+export const MajorMessage: React.FC<MajorMessageProps> = ({
```
vonElfvin, line 62: Couldn't this one just be called `Message` or is that too vague maybe?

src/components/member/tabs/DetailsTab.js
```diff
@@ -85,7 +86,7 @@ const DetailsTabComponent = (props) => {
   delete memberInfoWithoutSsn.__typename
 
   return memberInfoWithoutSsn ? (
-    <>
+    <EaseIn>
```
vonElfvin, line 89: Now that's nice, `<>` to `<EaseIn>` is something that is low effort to to going forward with a lot of feeling value

shared/hedvig-ui/animations/major-message.tsx
```diff
@@ -0,0 +1,91 @@
...
+    from: { opacity: 0, transform: 'translateY(2%)' },
+    to: { opacity: max, transform: 'translateY(0)' },
+  })
+
+const MajorMessageWrapper = styled('div')<{
+  paddingTop?: string
+  paddingBottom?: string
+  paddingLeft?: string
+  paddingRight?: string
+}>(({ paddingTop = '25vh', paddingBottom, paddingLeft, paddingRight }) => ({
```
vonElfvin, line 15: Correct me if I'm wrong here but it feels we don't use all of these props. I say we should keep it as slim as possible for now so it's easily extended for future needs instead? Challenge that if you want :)

vonElfvin (commented):
Now this is some R E M A R K A B L E work you've done here. Some minor comments to consider to make the code a bit more crispy. Great job!

palmenhq (commented):
SO! MANY! IMPROVEMENTS! :D 

Just an idea.. Could we use the Spinner (in the brilliantly named [`sipnner.ts`](https://github.com/HedvigInsurance/hedgehog/blob/master/shared/hedvig-ui/sipnner.tsx)  file 😬 ) as part of a loading state rather than text ellipsis? If not i think we should use the ellipsis animation everywhere to get even more consistency.

Another thought. What about making a dedicated `<Loader />` component or something similar just to not have to write `Loading` everywhere? Then also if we want to replace it with something completely different and even neater we could without much hassle

PR #249, opened 2020-10-29
Fade in lists one by one

vonElfvin:
Kapture 2020-10-29 at 20 54 13
Kapture 2020-10-29 at 20 59 53

### DIFF
--- a/shared/hedvig-ui/animations/fade-in.tsx
+++ b/shared/hedvig-ui/animations/fade-in.tsx
@@ -1,16 +1,22 @@
-import styled, { keyframes } from 'react-emotion'
+import styled, { keyframes, StyledComponent } from 'react-emotion'
 
-const fadeIn = (max) =>
+const fadeInKeyframes = (max) =>
   keyframes({
     from: { opacity: 0, transform: 'translateY(2%)' },
     to: { opacity: max, transform: 'translateY(0)' },
   })
 
-export const FadeInComponent = (component: any) =>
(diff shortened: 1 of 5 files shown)


rasmusguterstam (approved):
Let's goo 🔥

shared/hedvig-ui/animations/fade-in.tsx
```diff
@@ -6,10 +6,11 @@ const fadeIn = (max) =>
     to: { opacity: max, transform: 'translateY(0)' },
   })
 
-export const FadeIn = styled('div')<{ delay?: string }>(
-  ({ delay = '0ms' }) => ({
+export const FadeInComponent = (component: any) =>
```
palmenhq, line 9: Super nice and reusable! But a) i'd call this maybe something like `withFadeIn` or maybe actually just `fadeIn` (and rename the keyframes to `fadeInKeyframes` or so), the capital F indicates it'd be a component itsself which is not true - it's just a decorator :) and b) the typing needs some work.

Re the typing: Maybe we should try to "decorate" the styled function a bit more extensively. I think that'd be easiest considering emotions' typings being quite advanced. But i guess you could do something like:

```ts
const fadeIn: CreateStyledFunction<object> = (component, ...args) => styled(component, ...args)<{ delay?: string }>(({ delay = '0ms' }) => ({ 
...
```

palmenhq (commented):
awesome!

palmenhq (approved):
awesome! 👑

PR #251, opened 2020-11-13
show if member is not sign on unsigning

FredrikAreschoug:


styrelius (approved):
👍

src/features/tools/staging-tools/unsign-member-tool/index.tsx
```diff
@@ -40,11 +40,18 @@ const UnsignMemberToolComponent: React.FC<{} & WithShowNotification> = ({
...
+                  header: 'Success',
+                  message: `Successfully unsigned member`,
+                })
+
               showNotification({
                 type: 'olive',
                 header: 'Success',
-                message: `Successfully unsigned member`,
+                message: `Member was not signed!`,
               })
```
palmenhq, line 55: I haven't tried this myself so i might be wrong but since we don't make any returns and don't have any ifs.. won't this say both `Sucessfully unsigned member` and `Member was not signed!` at the same time?

palmenhq: As in... isn't the "real" issue that we end up in the `catch` branch if the member is unsigned because we're getting an internal server error, or is that maybe resolved?

palmenhq (commented):
Awesome and welcome to hedgehog! just a question really, i might be completely off here

src/features/tools/staging-tools/unsign-member-tool/index.tsx
AlexSJ1990: the internal-server error was solved in this pr -> https://github.com/HedvigInsurance/member-registry-service/pull/426/files so that it returns false rather than throwing. 

So thinking was here that if the result is true we will put "Successfully signed" and if it returns false "Member was not signed"

palmenhq: Aha right! But then i think we should put it like this insead:

```ts
if (result) {
  showNotification({
    type: 'olive',
    header: 'Success',
    message: `Successfully unsigned member`,
  })
} else {
  showNotification({
    type: 'olive',
    header: 'Success',
    message: `Member was not signed!`,
  })}
```

Trying to translate the current implementation to if statements i'd intpret it as something like this:

```ts
if (result) {
  showNotification({
    type: 'olive',
    header: 'Success',
    message: `Successfully unsigned member`,
  })

  // no early return or anything, just keep going
}
showNotification({
  type: 'olive',
  header: 'Success',
  message: `Member was not signed!`,
})
```

or am i missing something?

FredrikAreschoug: @palmenhq the if-statement was my first implementation but I got a tip that you cool typescript developers like early returns. But nice I will change it back then!

palmenhq: Haha, occasionally we don't dislike ifs :D i just started toying with codefresh in this repo just fyi, if CI fails its because of that. I can ping you when done

palmenhq (dismissed):
nice :D

palmenhq (commented):
Actually one thing i realised, maybe we could update the [changelog](https://github.com/HedvigInsurance/hedgehog/blob/master/src/changelog.ts)?

palmenhq:
@FredrikAreschoug feel free to merge in master now, codefresh should start worknig 👍

src/changelog.ts
```diff
@@ -11,6 +11,12 @@ export interface Change {
 }
 
 export const changelog: ReadonlyArray<Change> = [
+  {
+    date: '2020-11-17',
+    change:
+      'Unsigne tool now notifies if the member was signed or not on unsigning',
```
palmenhq, line 17: Typo :) (Unsigne -> Unsign)

palmenhq (approved):

PR #255, opened 2020-11-19
Add Monthly Entry functionality, with some refactoring

vonElfvin:
For added context:

**What is this?**
Monthly entries is a simple solution for IEX to not have to add object insurance manually each week. Currently it's very locked in to only support object insurance and nothing else (with source, external id etc. very locked in so they use it properly initially). It could potentially be expanded upon later to be used for more monthly entries though.

image

image

Kapture 2020-11-20 at 11 04 18


palmenhq:
@vonElfvin merge master and CI should start :)

schema.graphql
```diff
@@ -850,6 +854,19 @@ type Account {
...
+type MonthlyEntry {
+  id: ID!
+  externalId: String
+  amount: MonetaryAmountV2!
+  type: AccountEntryType!
+  source: String!
+  addedBy: String!
+  addedAt: Instant!
+  title: String!
+  comment: String!
```
palmenhq, line 869: I guess this is maybe a bit OT but should comment really be required?

shared/hedvig-ui/form.tsx
```diff
@@ -0,0 +1,171 @@
...
+import styled from 'react-emotion'
+import { Controller, useFormContext, ValidationRules } from 'react-hook-form'
+import {
+  Dropdown,
+  DropdownItemProps,
+  Form as SemanticForm,
+  FormProps,
+} from 'semantic-ui-react'
+
+export const Form: React.FC<{ onSubmit: (data) => void } & FormProps> = ({
```
palmenhq, line 13: can we add a type on `data` here?

shared/hedvig-ui/form.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+const ErrorMessageWrapper = styled('div')`
+  color: ${({ theme }) => theme.danger};
+  margin: 0;
+`
+
+const FormLabel: React.FC<{
+  label: string
+  name: string
+}> = ({ label, name }) => {
```
palmenhq, line 39: just an idea, shall we forward other props on the label as well? If you'd want to add something like idk.. onclick or styling or something? And maybe we could use `children`, it's kind of exactly this use case its for. So:

```ts
cost formLabel: React.FC<{ name: string } & React.HTMLProps<HTMLLabelElement>> = ({ name, children, ...props }) => {
  return <label htmlFor={name}{...props}>{children}</label>
}
```

I guess this applies to more components in this file really

shared/hedvig-ui/form.tsx
```diff
@@ -0,0 +1,171 @@
...
+}) => {
+  return (
+    <FormField {...props}>
+      <FormDropdownComponent options={options} {...props} />
+    </FormField>
+  )
+}
+
+interface FormFieldProps {
+  label: string
```
palmenhq: if we use `children` instead of label we could make this into a `React.ReactNode` and then we could pass any valid jsx in here, for example if you'd want a number, conditionals or a `span` tag :D

shared/hedvig-ui/form.tsx
```diff
@@ -0,0 +1,171 @@
...
+    >
+      <FormLabel label={label} name={name} />
+      {children}
+      <FormError name={name} />
+    </SemanticForm.Field>
+  )
+}
+
+interface SubmitButtonProps extends ButtonProps {
+  submitText: string
```
palmenhq: Suggestion: `children`

src/components/dashboard/index.tsx
```diff
@@ -60,13 +57,13 @@ export const Dashboard: React.FC<{ auth: any }> = ({ auth }) => {
 
   return (
     <Wrapper>
-      <Headline>
+      <MainHeadline style={{ marginBottom: '2rem' }}>
```
palmenhq, line 60: but... why

src/components/member/tabs/account-tab/AccountEntriesInfo.tsx
```diff
@@ -0,0 +1,19 @@
+import { FourthLevelHeadline, MainHeadline } from 'hedvig-ui/typography'
+import React from 'react'
+import styled from 'react-emotion'
+
+const AccountEntriesInfoWrapper = styled('div')`
+  margin-top: 4rem;
```
palmenhq, line 6: Imo we shouldn't add spacing around "encapsulated" components like this. Perhaps we could use the `Spacing` component if we want spacing around it where we use it, instead?

src/components/member/tabs/account-tab/AccountEntryTable.tsx
```diff
@@ -87,6 +87,15 @@ export const AccountEntryTable: React.FC<{
                         {entry.source}
                       </Grid.Column>
                     </Grid.Row>
+                    {entry.comment && (
+                      <Grid.Row>
+                        <Grid.Column>
+                          <span style={{ fontWeight: 'bold' }}>Comment</span>
```
palmenhq, line 93: Imo we shouldn't have inline styles, and especially not when we need to do it multiple times. Can we make a `Bold` component maybe?

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+const AddMonthlyEntryFormComponent: React.FC<{
+  memberId: string
+  preferredCurrency?: string
+} & WithShowNotification> = ({
+  memberId,
+  preferredCurrency,
+  showNotification,
+}) => {
+  if (!Boolean(preferredCurrency)) {
```
palmenhq, line 21: This is not wrong but you could just do `!preferredCurrency` :)

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+    return null
+  }
+
+  const form = useForm()
+
+  const [addMonthlyEntry] = useAddMonthlyEntry()
+  const onSubmit = (data) => {
```
palmenhq, line 28: type?

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const dataCopy = { ...data }
+    dataCopy.amount.currency = preferredCurrency
```
palmenhq, line 30: Nice! however this will only make a "shallow" copy so `dataCopy.amount.currency = preferredCurrency` will still modify the first object because it references the `amount` object which is present in both object. To prevent this you can do:

```ts
dataCopy.amount = { ...dataCopy.amount, currency: preferredCurrency }
```

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+    dataCopy.externalId =
+      dataCopy.externalId === '' ? undefined : dataCopy.externalId
```
palmenhq, line 32: just an idea, should we `trim` it to remove any whitespace  before making the empty check?

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+  return (
+    <FormProvider {...form}>
+      <Form onSubmit={onSubmit}>
+        <FormInput
+          label="External Id"
+          name="externalId"
+          defaultValue=""
+          rules={{
+            pattern: {
+              value: /^\d+$/,
```
palmenhq, line 60: just an idea, should we add type="number" to the input? And maybe to other number inputs in this form as well. Also, what about decimals, should you be able to put that or nah?

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+            required: 'Amount is required',
+          }}
+        />
+        <FormInput
+          label="Source"
+          name="source"
+          defaultValue="object"
+          rules={{
+            pattern: {
+              value: /^object$/,
```
palmenhq, line 88: soo... the only value you can put here is literally the string "object"? Seems unnecessary to even show this field to me if so. Maybe we should just name this form "add object insurance" and not show the input fields you can actually edit in that case

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+          rules={{ required: 'Comment is required' }}
+        />
+        <FormDropdown
+          options={enumToDropdownOptions(AccountEntryType)}
+          defaultValue={AccountEntryType.Subscription}
+          label="Type"
+          name="type"
+          rules={{
+            pattern: {
+              value: /^SUBSCRIPTION$/,
```
palmenhq, line 113: and same here, can you only put the string litteral "SUBSCRIPTION"? Maybe we should just name the form "add subscription entry" and not present the user with the option

src/components/member/tabs/account-tab/MonthlyEntriesTable.tsx
```diff
@@ -0,0 +1,91 @@
...
+                <Capitalized>{monthlyEntry.type}</Capitalized>
+              </Table.Cell>
+              <Table.Cell>{formatMoney(monthlyEntry.amount)}</Table.Cell>
+              <Table.Cell textAlign="center">
+                <Popover
+                  contents={
+                    <Grid>
+                      <Grid.Row>
+                        <Grid.Column>
+                          <span style={{ fontWeight: 'bold' }}>ID</span>
```
palmenhq, line 50: plox no inline styles, especially when repeated like this 😢

src/components/member/tabs/account-tab/index.tsx
```diff
@@ -83,44 +90,50 @@ export const AccountTab: React.FC<{
...
-                  account?.chargeEstimation?.discountCodes
-                )}
+                {formatMoney(account?.chargeEstimation.discount, moneyOptions)}
               </InfoText>
             </InfoRow>
             <Spacing top={'small'} />
+            {account?.chargeEstimation?.discountCodes.length > 0 && (
+              <InfoRow>
+                Discount References
+                <InfoText>{account?.chargeEstimation?.discountCodes}</InfoText>
```
palmenhq, line 118: does this work? shall we maybe use something like `account?.chargeEstimation?.discountCodes.join(', ')`? Just to not show `A_DISCOUNTANOTHER_DISCOUNT` and so

src/components/member/tabs/account-tab/index.tsx
```diff
@@ -83,44 +90,50 @@ export const AccountTab: React.FC<{
...
+                = {formatMoney(account?.chargeEstimation?.charge, moneyOptions)}
               </InfoText>
             </InfoRow>
           </InfoContainer>
         </Card>
-        <Card span={1} style={{ padding: '0.2rem' }}>
+      </CardsWrapper>
+      <AccountEntriesInfo />
+      <CardsWrapper>
+        <Card span={1} style={{ padding: '0.2rem', marginTop: '1.5rem' }}>
```
palmenhq, line 133: named component with proper styles plz :D

palmenhq (commented):
Great! Some input though :)

shared/hedvig-ui/form.tsx
vonElfvin: Yes in this case it should, since it's a monthly entry and thus a bit important to be documented as to why it exists

schema.graphql
vonElfvin: Yes, since it's an entry that will be added every month it's good to have the context for its existance being required.

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
vonElfvin: For now the external IDs for object insurnace are only integers tracked in an excel. Thus we only support integers.

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
vonElfvin: Other sources than object will be supported going forward. This is just to lock in the usage so IEX don't make typos etc and get to learn to tool "safely"

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
vonElfvin: As mentioned previously, more things will be supported going forward.

src/components/member/tabs/account-tab/index.tsx
vonElfvin: I guess? I didn't change the code just moved it. But I can investigate

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
palmenhq: Aha cool!

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
palmenhq: Right, clever! What about having a  dropdown instead then?

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
palmenhq: then the same counter-question :D dropdown?

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
vonElfvin: That is actually surprisingly smart. That way IEX don't have to remember all legible sources.

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
vonElfvin: I'll make a custom made one, good idea!

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
palmenhq: Haha! Surprisingly, i have my moments too 😉

vonElfvin: Haha it was not surprising that you came up with it, rather that a dropdown could be that good. Honestly we should use dropdowns more often for "structured strings" that IEX need to write. Thus we can also get around them making typos. Going forward you could even store the options in the database somehow so they can add choices themselves.

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  const form = useForm()
+
+  const [addMonthlyEntry] = useAddMonthlyEntry()
+  const onSubmit = (data: FieldValues) => {
+    const dataCopy = { ...data }
+    dataCopy.amount = { ...dataCopy.amount, currency: preferredCurrency }
+    dataCopy.externalId =
+      dataCopy.externalId.trim() === '' ? undefined : dataCopy.externalId
```
palmenhq, line 33: Maybe `dataCopy.externalId.trim() === '' ? undefined : dataCopy.externalId.trim()`? or assign it to a varible

codefresh.yml
```diff
@@ -1,65 +1,65 @@
-version: "1.0"
+version: '1.0'
```
palmenhq, line 1: 🤷

palmenhq (approved):
Awesome!

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
vonElfvin: I trim it in the backend instead 🧠 That way I can trim the title and comment as well in a nice way :)

PR #258, opened 2020-11-20
remove missed payments as options

AlexSJ1990:
we should only use this termination state when member has been terminated automatically due to failed payments - so we should remove the option from here


palmenhq (commented):
Did you do this by running `graphql:gen`? Otherwise please do that since this file shouldn't be edited manually :)

PR #260, opened 2020-11-20
remove missed_payments from termination reason

AlexSJ1990:
[ ] `src/changelog.ts` updated


schema.graphql
```diff
@@ -4,7 +4,6 @@ schema {
 }
 
 type QueryType {
-  monthlyPayments(month: YearMonth!): [MonthlySubscription]
```
palmenhq, line 7: is this maybe  some kind of merge conflict issue?

palmenhq (approved):
Hehe this is more what i expected :D

src/api/generated/graphql.tsx
```diff
@@ -1510,8 +1539,6 @@ export enum TerminationReason {
   MovedAbroad = 'MOVED_ABROAD',
   MovedInWithParents = 'MOVED_IN_WITH_PARENTS',
   Price = 'PRICE',
-  MissedPayments = 'MISSED_PAYMENTS',
-  MissedPaymentsBadRisk = 'MISSED_PAYMENTS_BAD_RISK',
```
vonElfvin, line 1514: I don't think removing it like this is a good approach. Since the termination reason might be sent from the backend. I'd say we'd rather send termination reason as a string and create an enum in the frontend of available choices.

palmenhq:
@AlexSJ1990 please delete the branch also unless there's a reason for keeping it :)

PR #261, opened 2020-11-24
Add ability to remove monthly entries

vonElfvin:
image

### DIFF
--- a/src/changelog.ts
+++ b/src/changelog.ts
@@ -12,9 +12,14 @@ export interface Change {
 
 export const changelog: ReadonlyArray<Change> = [
   {
+    date: '2020-11-24',
+    change: 'Add ability to remove a monthly entry',
+    authorGithubHandle: 'vonElfvin',
+  },
+  {
     date: '2020-11-23',
     change:
-      'Add ability "monthly entries" to a member\'s account, to be used to automate adding object insurances each month',
+      'Add ability "monthly entries" to a member\'s account, to be used to automate object insurances',
     authorGithubHandle: 'vonElfvin',
   },
   {
--- a/src/changelog.ts
+++ b/src/changelog.ts
@@ -13,7 +13,7 @@ export interface Change {
 export const changelog: ReadonlyArray<Change> = [
   {
     date: '2020-11-24',
-    change: 'Add ability to remove a monthly entry',
+    change: 'Add ability to remove a monthly entry from a member',
     authorGithubHandle: 'vonElfvin',
   },
   {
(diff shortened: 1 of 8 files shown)


src/changelog.ts
```diff
@@ -11,10 +11,15 @@ export interface Change {
 }
 
 export const changelog: ReadonlyArray<Change> = [
+  {
+    date: '2020-11-24',
+    change: 'Add ability to remove a monthly entry',
```
palmenhq, line 16: `..from  the members account` or something maybe?

src/components/member/tabs/account-tab/MonthlyEntriesTable.tsx
```diff
@@ -83,13 +95,48 @@ export const MonthlyEntriesTable: React.FC<{
...
                 >
-                  <DetailsIcon name="info circle" />
+                  <InfoCircleFill />
                 </Popover>
               </Table.Cell>
+              <Table.Cell>
+                <IconButton
+                  onClick={() => {
+                    const confirm = window.confirm(
+                      `Are you sure you want delete the monthly entry "${monthlyEntry.title}? (id=${monthlyEntry.id})"`,
```
palmenhq, line 105: very nit perhaps but these quotes look a bit weird to me?

```
Are you sure you want delete the monthly entry "the title? (id=1234)"
```

src/components/member/tabs/account-tab/MonthlyEntriesTable.tsx
```diff
@@ -83,13 +95,48 @@ export const MonthlyEntriesTable: React.FC<{
...
+                    )
+                    if (confirm) {
```
palmenhq, line 107: Just to decrease the indentation level here i think an early return would be nicer :)
```ts
if (!confirm('u sure')) {
  return
}

**Review by palmenhq — COMMENTED** [review]
Code-wise it looks good :)
Two general questions:
- are we implicitly making "undo" functionality here (if they accidentally added a monthly entry)? If so maybe we should also create an undo button instead. I'm just guessing though, since this review was needed soon.
- Should one be able to  "stop" the recurring entries in the future, or is now now now enough? What happens if someone cancels their object insurance mid-month, for example?

File: src/components/member/tabs/account-tab/MonthlyEntriesTable.tsx [hunk 4]
```diff
@@ -83,13 +95,49 @@ export const MonthlyEntriesTable: React.FC<{
...
                 >
-                  <DetailsIcon name="info circle" />
+                  <InfoCircleFill />
                 </Popover>
               </Table.Cell>
+              <Table.Cell>
+                <IconButton
+                  onClick={() => {
+                    const confirm = window.confirm(
+                      `Are you sure you want delete the monthly entry titled "${monthlyEntry.title}? (id=${monthlyEntry.id})"`,
```
palmenhq, line 105: What i ment here was that the quotes look a bit off. Probably the ending quote should be before the question mark right?

palmenhq (approved):
Looks great 👍

PR #262, opened 2020-11-25
Feature/itemizer improvements

rasmusguterstam:
✅ `src/changelog.ts` updated

### What's up with itemizer

- Provides total valuation and deduction for entire inventory
- Explicit explanation for each item valuation (Custom valuation, Market price, No depreciation (100%) or Depreciation)
- Plenty of refactoring (although still pretty wack code, but substantially better) 😩
 

image


schema.graphql
```diff
@@ -19,7 +19,11 @@ type QueryType {
   findPartnerCampaigns(input: CampaignFilter!): [VoucherCampaign!]!
   getPartnerCampaignOwners: [CampaignOwnerPartner!]!
   dashboardNumbers: DashboardNumbers
-  getClaimItemValuation(request: GetValuationInput): ClaimItemValuation!
+  getClaimItemValuation(
+    request: GetClaimItemValuationInput
+  ): ClaimItemValuation!
+  getClaimValuation(claimId: ID!, typeOfContract: String): ClaimValuation!
```
palmenhq, line 25: is the `typeOfContract` nullable here by accident or intention?

src/components/claims/claim-details/components/claim-items/chips/Chips.tsx
```diff
@@ -0,0 +1,56 @@
+import { NoValuationChip } from 'components/claims/claim-details/components/claim-items/chips/components/NoValuationChip'
+import { MessageChip } from 'components/claims/claim-details/components/claim-items/chips/MessageChip'
```
palmenhq, line 2: maybe we could shorten these to just `./TheComponent` so its clear they belong together?

src/components/claims/claim-details/components/claim-items/chips/Chips.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { UpsertClaimItemVariables } from 'graphql/use-upsert-claim-item'
+import React from 'react'
+import { CustomValuationChip } from './CustomValuationChip'
+
+export const Chips: React.FC<{
+  typeOfContract: string
+  formData: UpsertClaimItemVariables
+  setFormData: (formData: UpsertClaimItemVariables) => void
+}> = ({ typeOfContract, formData, setFormData }) => {
+  const { itemFamilyId, itemTypeId } = { ...formData }
```
palmenhq, line 13: no need to copy the form data here actually :) you could just do

```ts
const { itemFamilyId, itemTypeId } = formData
 ```

File: src/components/claims/claim-details/components/claim-items/chips/CustomValuationChip.tsx [hunk 4]
```diff
@@ -0,0 +1,30 @@
...
+import { InputChip } from './components/InputChip'
+
+export const CustomValuationChip: React.FC<{
+  formData: UpsertClaimItemVariables
+  onChange: (amount: string) => void
+  onReset: () => void
+}> = ({ formData, onChange, onReset }) => {
+  const { customValuationAmount: amount, purchasePriceCurrency: currency } = {
+    ...formData,
+  }
```
palmenhq, line 13: same here no real need to copy the `formData` here

palmenhq, line 13: Just an idea, what about passing only these variables instead of the whole `formData` object to make it clearer what is the actual dependency?

src/components/claims/claim-details/components/claim-items/chips/MessageChip.tsx
```diff
@@ -0,0 +1,68 @@
...
+  })
+
+  React.useEffect(() => {
+    if (claimItemValuation) {
+      onValuation(claimItemValuation)
+    }
+  }, [claimItemValuation])
+
+  const { depreciatedValue, valuationRule } = { ...claimItemValuation }
+  const { valuationType } = { ...valuationRule }
```
palmenhq, line 44: no need to copy :D

src/components/claims/claim-details/components/claim-items/index.tsx
```diff
@@ -1,26 +1,126 @@
-import { Grid } from '@material-ui/core'
+import { Button } from '@material-ui/core'
 import { Contract } from 'api/generated/graphql'
+import { Chips } from 'components/claims/claim-details/components/claim-items/chips/Chips'
+import { TotalValuationChip } from 'components/claims/claim-details/components/claim-items/chips/components/TotalValuationChip'
+import {
+  isEmpty,
+  isValidDate,
+  isValidNumber,
+} from 'components/claims/claim-details/components/claim-items/utils'
```
palmenhq, line 9: maybe we could import these like `./components`?

src/components/claims/claim-details/components/claim-items/index.tsx
```diff
@@ -1,26 +1,126 @@
...
+import { ItemList } from './item-list/ItemList'
+
+const BottomWrapper = styled.div`
+  display: flex;
+  justify-content: space-between;
+`
+
+const ChipsWrapper = styled.div``
+
+const formLooksGood = (request: UpsertClaimItemVariables) => {
```
palmenhq, line 31: maybe `formIsValid` would be a more describing name?

src/components/claims/claim-details/components/claim-items/item-form/ItemForm.tsx
```diff
@@ -0,0 +1,165 @@
...
+            placeholder="Purchase date"
+            inputProps={{
+              style: {
+                paddingLeft: '10px',
+                paddingRight: '10px',
+              },
+            }}
+          />
+        </Grid>
+        <Grid item xs={true}>
```
palmenhq, line 146: no need for `={true}` here actually  :)

src/components/claims/claim-details/components/claim-items/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { isAfter, isValid, parseISO } from 'date-fns'
+
+export const isValidDate = (date?: string) =>
+  !!date &&
+  (date === ''
+    ? true
+    : isValid(parseISO(date)) && !isAfter(parseISO(date), new Date()))
```
palmenhq, line 7: since `!!date` will always be false if `date` is `''` (empty string) i guess the first two checks aren't really needed

src/components/claims/claim-details/index.tsx
```diff
@@ -125,10 +116,19 @@ const ClaimPage: React.FC<Props> = ({ ...props }) => (
                     <Grid item xs={12}>
                       <ClaimItems
                         claimId={props.match.params.claimId}
-                        memberId={member?.memberId ?? null}
-                        contract={contract}
+                        memberId={props.match.params.memberId}
```
palmenhq, line 119: why? Isn't it better to get the "real" member id than the member id from the url?

palmenhq (commented):
I didn't have the time currently to check too thoroughly but looks awesome! I'll have a closer look asap but good job with this :)

schema.graphql
rasmusguterstam: Haha good to always verify this - but in this specific case it is actually intentional! You can get the total valuation of a claim without typeOfContract, but you can't get the associated deductible!

src/components/claims/claim-details/components/claim-items/chips/MessageChip.tsx
rasmusguterstam: You were right in the other scenarios, but in this specific case it actually solves the problem of `Property 'depreciatedValue' does not exist on type 'ClaimItemValuation | undefined'`, since spreading `undefined` gives an object of nothing (which at least is an object, which we can _try_ to destructure, even though we know it would turn out undefined).

Thought this was a neat solution, but perhaps there's a better way to do this? 🧐

src/components/claims/claim-details/index.tsx
rasmusguterstam: Haha good point, didn't really think of the origin of the variable, just thought I would skip the chaining and undefined-check 🤦‍♂️

src/components/claims/claim-details/components/claim-items/utils.ts
rasmusguterstam: Turns out it wasn't even used at all 😅

palmenhq (dismissed):
amazee

src/components/claims/claim-details/components/claim-items/index.tsx
```diff
@@ -1,26 +1,122 @@
...
+  memberId: string
+  contract: Contract | null
 }> = ({ claimId, memberId, contract }) => {
+  const [claimItems] = useGetClaimItems(claimId)
+  const [contractMarketInfo] = useContractMarketInfo(memberId)
+  const { typeOfContract } = { ...contract }
+  const [claimValuation] = useGetClaimValuation(claimId, typeOfContract)
+  const [upsertClaimItem, { loading }] = useUpsertClaimItem(claimId)
+
+  const { preferredCurrency = 'SEK' } = { ...contractMarketInfo }
```
vonElfvin, line 66: No defaulting to 'SEK' pls :)
```suggestion
  const { preferredCurrency } = { ...contractMarketInfo }
```

src/components/claims/claim-details/components/claim-items/chips/Chips.tsx
```diff
@@ -0,0 +1,57 @@
...
+    itemTypeId,
+    typeOfContract,
+  )
+
+  React.useEffect(() => {
+    setFormData({ ...formData, customValuationAmount: '' })
+  }, [itemFamilyId, itemTypeId])
+
+  if (loading || !itemFamilyId || !typeOfContract) {
+    return <></>
```
vonElfvin, line 26: We usually return null instead in these cases

vonElfvin (approved):

PR #264, opened 2020-11-30
Package upgrades

palmenhq:
<strike>[ ] `src/changelog.ts` updated</strike>

Casually patchin' up a good 27 509 vulnerabilities. We'll try to do this more often moving forward so the upgrade process doesn't have to be so big.


gustaveen (approved):
Looks good to me 🙌

AlexSJ1990 (approved):
🧹

vonElfvin (approved):
Nice!

PR #265, opened 2020-12-02
Use react-hook-form for Questions and Account Entry

vonElfvin:
From:
image

To:
image

From:
image

To:
image
image

Icons and affix to the input fields:
image


src/components/member/tabs/account-tab/AddEntryForm.tsx
```diff
@@ -1,215 +1,202 @@
...
+    text: 'Other',
+  },
+]
 
 const AddEntryFormComponent: React.FC<{
   memberId: string
 } & WithShowNotification> = ({ memberId, showNotification }) => {
   const [contractMarketInfo] = useContractMarketInfo(memberId)
-  const preferredCurrency = contractMarketInfo?.preferredCurrency || 'SEK'
+  const preferredCurrency = contractMarketInfo!!.preferredCurrency
```
rasmusguterstam, line 90: Do we always have contractMarketInfo??

src/components/member/tabs/account-tab/AddMonthlyEntryForm.tsx
```diff
@@ -20,7 +21,11 @@ const AddMonthlyEntryFormComponent: React.FC<{
   showNotification,
 }) => {
   if (!preferredCurrency) {
-    return null
+    return (
+      <StandaloneMessage>
+        The member has no preferred currency
+      </StandaloneMessage>
+    )
   }
```
rasmusguterstam, line 29: 💅

rasmusguterstam (dismissed):

src/components/questions/questions-list/AnswerForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+        showNotification({
+          type: 'red',
+          header: `Unable to send the answer to ${memberId}`,
+          message: error.message,
+        })
+        throw error
+      })
+  }
+
+  const doneClick = () => {
```
rasmusguterstam, line 71:
```suggestion
  const onDone = () => {
```

src/components/questions/questions-list/AnswerForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+          <Form onSubmit={onSubmit}>
+            <Grid container spacing={24}>
+              <Grid item xs={9}>
+                <FormTextArea
+                  name="answer"
+                  defaultValue=""
+                  rules={{
+                    required: 'Cannot send an empty message',
+                    pattern: {
+                      value: /[^\s]/,
```
palmenhq, line 103: this regex is a bit confusing to me, maybe we could do `/^.+$/` and always trim it before validation or something? Optinaly have two rules if that's possible?

src/components/questions/questions-list/AnswerForm.tsx
palmenhq: I'd even suggest `handleDone`, as this is the handler, not the listener

palmenhq (commented):
Really awesome, just a couple of minor things!

src/components/questions/questions-list/AnswerForm.tsx
vonElfvin: Since we trim it before sending I changed to a regex that just makes sure it contains at least one character or number :) `/[a-zA-Z0-9]+/`

vonElfvin: I agree that the regex was confusing, I think this is a lot more explicit

vonElfvin: Actually, this is the only regex that does the job without trimming. I think it's decently explicit, "at least one character that is not a space"

PR #270, opened 2021-01-07
Chore: Kill claims store

rasmusguterstam:
## What

- Kills off Claims Store and implements GraphQL instead
- Removes old, unused components
- Removes "BackendServed{blahblah}" and implements a single `Paginator` instead. There seemed to be two different paginators: one for paginating locally stored arrays, and one for paginating directly from back-office. The former was never used anywhere and just made it confusing.
- Refactored "claims tab" under "member"

### Member claims tab (before)
https://user-images.githubusercontent.com/51323202/103869420-8e8aa980-50ca-11eb-800e-533f154dbb81.mov

### Member claims tab (after)
https://user-images.githubusercontent.com/51323202/103869070-1a500600-50ca-11eb-839c-9b93779e077b.mov

### Claims list (before)
image

### Claims list (after)
image


src/components/member/tabs/ClaimsTab.tsx
```diff
@@ -1,202 +1,154 @@
...
+                variation="danger"
+                color="danger"
+                onClick={() => {
+                  setShowForm(false)
+                  setClaimDate(null)
+                  setClaimSource(null)
+                }}
+              >
+                Cancel
+              </Button>
```
vonElfvin, line 96: I would wrap the button with the spacing here.
```suggestion
              <Spacing left={'small'}>
              <Button
                disabled={createClaimLoading}
                variation="danger"
                color="danger"
                onClick={() => {
                  setShowForm(false)
                  setClaimDate(null)
                  setClaimSource(null)
                }}
              >
                Cancel
              </Button>
              </Spacing>
```

src/components/member/tabs/ClaimsTab.tsx
```diff
@@ -1,202 +1,154 @@
...
+              <Spacing left={'small'} />
+              <Button
```
vonElfvin, line 98: Same here

src/components/member/tabs/ClaimsTab.tsx
```diff
@@ -1,202 +1,154 @@
...
+                  claimSource === null ||
+                  claimDate === null ||
+                  createClaimLoading
+                }
+                variation="success"
+                color="success"
+                onClick={() => {
+                  if (claimSource === null || claimDate === null) {
+                    return
+                  }
```
vonElfvin, line 109: If they are null and the disabled doesn't do it's just we may as well crash in the backend instead and handle the `catch` below with some appropriate logging

src/components/member/tabs/ClaimsTab.tsx
```diff
@@ -1,202 +1,154 @@
...
+                    },
+                  }).then((response) => {
+                    const claimId = response.data?.createClaim
+                    setShowForm(false)
+                    setClaimDate(null)
+                    setClaimSource(null)
+                    if (claimId) {
+                      history.push(`/claims/${claimId}/members/${memberId}`)
+                    }
+                    refetch()
```
vonElfvin, line 125: Guess we can skip some steps if we get the claim id? With that said don't we always get the claim Id?
```suggestion
                    const claimId = response.data?.createClaim
                    if (claimId) {
                      history.push(`/claims/${claimId}/members/${memberId}`)
                      return
                    }
                    setShowForm(false)
                    setClaimDate(null)
                    setClaimSource(null)
                    refetch()
```

src/components/member/tabs/ClaimsTab.tsx
```diff
@@ -1,202 +1,154 @@
...
+      </HeaderWrapper>
+      <Table celled selectable>
+        <ClaimListHeader />
+
+        <Table.Body>
+          {claims.map((item, index) => (
+            <ClaimListItem key={item.id} item={item} index={index} />
+          ))}
+        </Table.Body>
+      </Table>
```
vonElfvin, line 151: Do we have a parent component for the other claims list we could just pass the claims to maybe?

src/components/shared/paginator/Paginator.tsx
```diff
@@ -0,0 +1,119 @@
+import { range } from 'lib/helpers'
+import React from 'react'
+import styled, { css } from 'react-emotion'
+import { Button, Table } from 'semantic-ui-react'
+
+interface BackendPageState {
```
vonElfvin, line 6:
```suggestion
interface PageState {
```

src/components/shared/paginator/Paginator.tsx
```diff
@@ -0,0 +1,119 @@
+import { range } from 'lib/helpers'
+import React from 'react'
+import styled, { css } from 'react-emotion'
+import { Button, Table } from 'semantic-ui-react'
```
vonElfvin, line 4: May be a lot to ask for, but maybe we can try to use hedvig-ui buttons
```suggestion
import { Button, ButtonGroup } from 'shared/hedvig-ui'
```

src/components/shared/paginator/Paginator.tsx
```diff
@@ -0,0 +1,119 @@
...
+                {itemContent(item, index)}
+              </React.Fragment>
+            ))}
+          </Table.Body>
+        ) : (
+          <></>
+        )}
+      </Table>
+      <PaginatorContainer>
+        {totalPages > 1 && (
```
vonElfvin, line 91: Guess we don't need the container if it's only one page
```suggestion
        {totalPages > 1 && (
         <PaginatorContainer>
```

src/containers/claims-page/ClaimsPage.tsx
```diff
@@ -1,19 +1,9 @@
...
+    <Claims />
   </ListPage>
 )
-
-export default withRouter(
-  // @ts-ignore
-  connect(({ claims }) => ({ claims }), {
-    ...actions.claimsActions,
-  })(ClaimsPage),
-)
```
vonElfvin, line 19: If this works, which I guess it does, we can do this on other places as well I think (places without any props/actions that then don't need the connect)

src/graphql/use-get-member-claims.tsx
```diff
@@ -0,0 +1,22 @@
+import {
+  Claim,
+  GetMemberClaimsQueryResult,
+  useGetMemberClaimsQuery,
+} from 'api/generated/graphql'
+
+type UseGetMemberClaimsReturnTuple = [
+  Claim[] | undefined,
```
vonElfvin, line 8:
```suggestion
  ReadonlyArray<Claim> | undefined,
```

src/components/claims/claims-list/components/ClaimListHeader.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react'
+import { Table } from 'semantic-ui-react'
+
+type SortDirection = 'ascending' | 'descending'
```
vonElfvin, line 4: We could create an enum here instead:
```suggestion
enum SortDirection {
  Ascending = 'ASC',
  Descending = 'DESC'
}
```

src/components/member/tabs/ClaimsTab.tsx
```diff
@@ -1,202 +1,154 @@
...
+                    return
+                  }
+
+                  createClaim({
+                    variables: {
+                      memberId,
+                      date: format(claimDate, "yyyy-MM-dd'T'HH:mm:ss"),
+                      source: claimSource,
+                    },
+                  }).then((response) => {
```
vonElfvin, line 117: Let's add a `catch` with some appropriate logging I'd say

src/components/member/tabs/ClaimsTab.tsx
vonElfvin: Do we even want the automatic navigation btw? Seems a bit weird to just fly away from the page, especially if it's not a 100% behaviour

vonElfvin (commented):
Now this is 🔥 Redux is almost gone! 🔜 ™️ A lot awesome bonus improvements as well 🧯

src/components/login/login-process/LoginProcess.js
```diff
@@ -1,25 +0,0 @@
-import PropTypes from 'prop-types'
-import React from 'react'
-import { Dimmer, Loader } from 'semantic-ui-react'
-
-export default class LoginProcess extends React.Component {
```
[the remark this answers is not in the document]
rasmusguterstam: 🤷‍♂️ Whatever it was, it is no more

src/components/shared/paginator/Paginator.tsx
rasmusguterstam: Good point!

src/containers/claims-page/ClaimsPage.tsx
rasmusguterstam: I actually think we've removed all the ones that we're not using anymore!

src/components/member/tabs/ClaimsTab.tsx
rasmusguterstam: The one for the claims list is based on the `Paginator` component which paginates from the back-end, but this one only lists all claims for a specific member (resolved from `Member`). I guess you could make a dynamic paginator which could manage both local and back-end-served pagination, but this seemed like the easy way! 😄

vonElfvin (dismissed):
Lit 🔥

vonElfvin (approved):

PR #272, opened 2021-01-18
Fix/enable dd generation link

vonElfvin:


meletios (approved):
Do we need to restrict that to Kalle only? Double check with Isabella!

meletios (approved):

PR #280, opened 2021-02-11
Add ability to dynamically filter number of teams in Questions tab

vonElfvin:
# Jira Issue: []

## What?
- Add ability to opt in for more colors in H.OPE.
- Add urls to member tabs for future tracking and keyboard navigation
- Add master inception and termination date to claims view to reduce jump to member page

## Why?
- Add opt-in for more colors if workload in H.OPE. is too big for 2 splits
- To speed up work in H.OPE. slightly

## Optional screenshots
image

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


src/components/questions/filter.tsx
```diff
@@ -1,18 +1,24 @@
...
+
 export enum FilterState {
-  Even,
-  Odd,
+  First,
+  Second,
+  Third,
   Sweden,
   Norway,
   Denmark,
```
palmenhq, line 24: I'm starting to think it doesn't make sense to put all of these filters within the same enum. Could we find a way of separating it maybe?

src/utils/member.tsx
```diff
@@ -21,28 +26,16 @@ export const MemberAge: React.FC<{
...
+  return `${FilterState[getColorNumberForMember(memberId, numberColors)]}`
 }
 
-export const getMemberGroup = (memberId: string) => {
-  if (isMemberIdEven(memberId.toString())) {
-    return 'Red team'
-  }
-  return 'Green team'
+export const memberBelongsToColor = (memberId, colorNumber, numberColors) => {
+  return getColorNumberForMember(memberId, numberColors) === colorNumber
```
palmenhq, line 38: Does this function make sense now that we've abstracted the `getMemberIdColor` to a nice function? Wouldn't it be even easier to just do the comparison inline?

palmenhq (commented):
Great stuff! 😍

What's your thoughts on clarifying what are these colors? Unless you're aware of the concept that `FilterState.First` maps to "team read", which is what IEX use to split workload, the word `color` is kinda generic. I liked `Team` more but maybe we could find something better.

src/utils/number-colors-context.tsx
```diff
@@ -1,25 +1,26 @@
...
     console.error(e)
     return 2
   }
 }
 
-export const NumberColorsContext = createContext({
+export const NumberColorsContext = createContext<{
+  numberColors: number
+  setNumberColors: (value: number) => void
+}>({
```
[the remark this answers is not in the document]
vonElfvin: I changed it to "MemberGroup"

src/components/questions/filter.tsx
vonElfvin: I agree, but I think the investment is too high given how short-lived this way of working will be.

src/components/claims/claim-details/components/MemberInformation.tsx
```diff
@@ -88,6 +93,22 @@ const MemberInformation: React.FC<{
...
+      </p>
+      {lastTermination && (
+        <p>
+          <strong>Last Termination Date:</strong> {lastTermination} (
+          {lastTermination &&
+            formatDistance(new Date(lastTermination), new Date(), {
+              addSuffix: true,
+            })}
+        </p>
+      )}
```
rasmusguterstam, line 111: I know there's a lot of places where we do this, but perhaps we should try to use `<Paragraph>` and `<ThirdLevelHeadline>` from `hedvig-ui` whenever we can? :)

src/components/member/tabs/quote-tab/index.tsx
```diff
@@ -98,7 +98,7 @@ export const Quotes: React.FunctionComponent<{ memberId: string }> = ({
 
   const getTabs = () =>
     getUniqueContractTypes().map((contractType) => ({
-      menuItem: getTextFromEnumValue(contractType),
+      menuItem: getTextFromEnumValue(contractType, true),
```
rasmusguterstam, line 101: I know that WebStorm hints on the arguments, but it looks a bit weird to not be able to specify the parameter explicitly (like `capitalized: true`). Not a biggie tho!

src/components/questions/filter.tsx
```diff
@@ -91,8 +105,33 @@ export const QuestionsFilter: React.FC<{
     return questionGroups.filter(filterer([filter])).length
   }
 
+  const { numberMemberGroups } = useContext(NumberMemberGroupsContext)
```
rasmusguterstam, line 108: How about we define a custom hook? 😎

Maybe something like:

```javascript

const { numberOfColors } = useNumberOfColors()

```

src/components/questions/filter.tsx
palmenhq: Is there a plan to stop using this way of working, and an alternative that we know of? Otherwise i think we'll be here for quite some time

src/utils/number-member-groups-context.tsx
```diff
@@ -0,0 +1,50 @@
...
+    )
+  }, [numberMemberGroups])
+
+  return (
+    <NumberMemberGroupsContext.Provider
+      value={{
+        numberMemberGroups,
+        setNumberMemberGroups: (newNumberMemberGroups: number) => {
+          setNumberMemberGroups(newNumberMemberGroups)
+        },
```
palmenhq, line 44: should work right?
```suggestion
        setNumberMemberGroups,
```

palmenhq (commented):
Muy bien!

src/utils/number-member-groups-context.tsx
```diff
@@ -0,0 +1,50 @@
...
+import React, { createContext, useEffect, useState } from 'react'
+
+const NUMBER_MEMBER_GROUPS_KEY = 'hedvig:member:groups:number'
+
+export const getDefaultNumberMemberGroups = (): number => {
+  try {
+    const numberOfMemberGroups = Number(
+      window.localStorage.getItem(NUMBER_MEMBER_GROUPS_KEY),
+    )
+    return Math.min(numberOfMemberGroups, totalNumberMemberGroups) ?? 2
```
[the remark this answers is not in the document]
vonElfvin: Oh yea nice catch, that's a mistake :)

src/components/member/tabs/quote-tab/index.tsx
vonElfvin: Don't think you can do that in TypeScript? I'd love to do it, but afaik you can only do it in Kotlin (among others ofc)

vonElfvin: We could do it if we put it as an object but I think that's a bit weird in this case

rasmusguterstam: Yes, that would be the alternative! But yeah, I agree that it seems overkill. With that said, it seems to pretty standard among JS libraries, so might be something to think about in future scenarios!

rasmusguterstam (approved):
🤘✅

PR #282, opened 2021-02-15
(Do not merge, will be replaced by CLI)Feature/keyboard navigation

vonElfvin:
# Jira Issue: Mini Hackathon project

## What?
Add keyboard navigation, including:
- Option-D: Dashborad
- Option-Q: Questions
- Option-T: Tools
- Option-C: Claims
- Option-S: Member Search
- Option-L: Logout

Also
- Option-A: Adyen
- Option-R: Trustly
- Option-G: GSR

And even:
- Option-Arrow Left/Right for member tab navigation
- Option-W for chat window resizing
- Option-M for member navigation in claims view
- Option-Backspace for most recently visited claim

## Why?
- Another step towards mouse-less H.OPE. ✊ 

## Optional screenshots
Navigation:
image
Claims page:
image
Chat:
image
image

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/src/components/shared/navigation/breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/shared/navigation/breadcrumbs/Breadcrumbs.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import styled from 'react-emotion'
 import { useHistory } from 'react-router'
 import { Link } from 'react-router-dom'
@@ -23,9 +23,11 @@ const Breadcrumbs: React.FC = () => {
     history.location.pathname,
   )
 
-  history.listen((location) => {
-    setPathname(location.pathname)
-  })
+  useEffect(() => {
+    history.listen((location) => {
+      setPathname(location.pathname)
+    })
+  }, [])
 
   if (pathname.startsWith('/login')) {
     return null
--- a/src/components/shared/navigation/sidebar/VerticalMenu.tsx
+++ b/src/components/shared/navigation/sidebar/VerticalMenu.tsx
@@ -1,7 +1,9 @@
 import { colorsV3 } from '@hedviginsurance/brand'
+import { Spacing } from 'hedvig-ui/spacing'
 import React, { useContext, useEffect, useState } from 'react'
 import {
   ArrowUpRight,
+  Backspace,
   BoxArrowLeft,
   ChevronLeft,
   CreditCard,
@@ -22,18 +24,8 @@ import { authLogOut, AuthState } from 'store/actions/auth'
 import { BackofficeStore } from 'store/storeTypes'
 import { DarkmodeContext } from 'utils/darkmode-context'
 import {
-  A_KEY_CODE,
-  BACKSPACE_KEY_CODE,
-  C_KEY_CODE,
-  D_KEY_CODE,
-  G_KEY_CODE,
-  L_KEY_CODE,
-  OPTION_KEY_CODE,
-  Q_KEY_CODE,
-  R_KEY_CODE,
-  S_KEY_CODE,
-  T_KEY_CODE,
-  useKeyPressed,
+  KeyCode,
+  useKeyIsPressed,
   usePressedKey,
 } from 'utils/hooks/key-press-hook'
 import { Logo, LogoIcon } from './elements'
@@ -236,6 +228,12 @@ const routes = {
   gsr: 'https://app.gsr.se/Account/SignIn',
 }
 
+interface LatestClaim {
+  memberId: string
+  claimId: string
+  location: string
+}
+
 export const VerticalMenuComponent: React.FC<any & { history: History }> = ({
   authLogOut: authLogOut_,
   loginState,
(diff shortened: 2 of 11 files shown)


src/utils/hooks/key-press-hook.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const G_KEY_CODE = 71
+export const L_KEY_CODE = 76
+export const C_KEY_CODE = 67
+export const M_KEY_CODE = 77
+export const LEFT_KEY_CODE = 37
+export const RIGHT_KEY_CODE = 39
+export const ONE_KEY_CODE = 49
+export const NINE_KEY_CODE = 57
+export const BACKSPACE_KEY_CODE = 8
+export const ENTER_KEY_CODE = 13
```
bystam, line 20: Would it make sense to group these under a `KeyCodes` object? With usage like `KeyCode.M`?

src/utils/hooks/key-press-hook.tsx
```diff
@@ -0,0 +1,89 @@
...
+      window.removeEventListener('keydown', handleKeydown)
+      window.removeEventListener('keyup', handleKeyup)
+      document.removeEventListener('visibilitychange', handleVisibility)
+    }
+  }, [])
+
+  return keyPressed
+}
+
+export const usePressedKey = (): number | null => {
```
bystam, line 60: It's not obvious to me what the difference between `usePressedKey` and `useKeyPressed` are based on their name. Would a doc comment help here maybe?

src/utils/hooks/key-press-hook.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const RIGHT_KEY_CODE = 39
+export const ONE_KEY_CODE = 49
+export const NINE_KEY_CODE = 57
+export const BACKSPACE_KEY_CODE = 8
+export const ENTER_KEY_CODE = 13
+
+export const isAllowedOptionKeyCode = (keyCode: number | null) =>
+  keyCode == null || (keyCode >= ONE_KEY_CODE && keyCode <= NINE_KEY_CODE)
+
+export const useKeyPressed = (keyCode: number): boolean => {
```
bystam, line 25: Based on the usage it looks like this one could be named `useKeyIsDown` maybe?

bystam (commented):
Veeeery cool 😍 I am amazed at the level of niceness Hope has!

src/utils/hooks/key-press-hook.tsx
vonElfvin: Changed to `keyIsPressed` since "down" is a bit taken by the default interface using "up" and "down"

src/utils/hooks/key-press-hook.tsx
vonElfvin: changed to `useKeyIsPressed`

src/components/shared/navigation/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -23,9 +23,11 @@ const Breadcrumbs: React.FC = () => {
...
   )
 
-  history.listen((location) => {
-    setPathname(location.pathname)
-  })
+  useEffect(() => {
+    history.listen((location) => {
+      setPathname(location.pathname)
+    })
+  }, [])
```
styrelius, line 30: I might be wrong but I think the same result is achieved by using the `useLocation` hook. And picking the `pathname` property from it.

src/components/shared/navigation/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -23,9 +23,11 @@ const Breadcrumbs: React.FC = () => {
     history.location.pathname,
```
styrelius, line 23: Sorry, I know this isn't a new change but the way I've understood it is that it's recommended that you use `useLocation` to get the current location, instead of picking the `locale` property from the history object with `useHistory` (has to do with the fact that the history object is mutable)

src/components/shared/navigation/sidebar/VerticalMenu.tsx
```diff
@@ -217,8 +242,87 @@ export const VerticalMenuComponent: React.FC<any & { history: History }> = ({
...
+  const [latestClaim, setLatestClaim] = useState<LatestClaim | null>(null)
   const { isDarkmode, setIsDarkmode } = useContext(DarkmodeContext)
+  const optionIsPressed = useKeyIsPressed(KeyCode.Option)
+  const pressedKey = usePressedKey()
+
+  React.useEffect(() => {
+    history.listen((location) => {
+      setCurrentLocation(location.pathname)
+    })
+  }, [])
```
styrelius, line 258: Same as the comment above ☝️

src/components/shared/navigation/sidebar/VerticalMenu.tsx
```diff
@@ -217,8 +242,87 @@ export const VerticalMenuComponent: React.FC<any & { history: History }> = ({
...
+
+  React.useEffect(() => {
+    const newLocations = [...locations]
+    newLocations.push(currentLocation)
+    if (newLocations.length > 10) {
+      newLocations.shift()
+    }
```
styrelius, line 265: OK feel very free to just ignore this since it's not specifically React related but would this make it so you don't have to reverse the array on line 271? You only get the last 10 locations, in the order you add them:
```suggestion
    const latestLocations = [currentLocation, ...locations]
    const newLocations = newLocations.filter((_, index) => index < 10)
```

styrelius (commented):
Really nice with keyboard shortcuts! 🙌 ✨  Hade a few comments, most of them regarding React Router implementation.

PR #283, opened 2021-02-24
Show carrier that applies to a claim on a claim

vonElfvin:
# Jira Issue: []

## What?
- Show the carrier on the claim

## Why?
- To ensure the receival of carrier works

## Optional screenshots
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -332,6 +332,7 @@ type GenericAgreement {
   extraBuildings: [ExtraBuilding!]
   isSubleted: Boolean
   lineOfBusinessName: String!
+  carrier: String!
 }
 
 enum SwedishApartmentLineOfBusiness {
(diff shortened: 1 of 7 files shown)


schema.graphql
```diff
@@ -937,25 +938,31 @@ enum ClaimSource {
...
+  state: ClaimState!
   type: ClaimType
   reserves: MonetaryAmount
-  registrationDate: Instant
-  notes: [ClaimNote]
-  transcriptions: [ClaimTranscription]
-  payments: [ClaimPayment]
-  events: [ClaimEvent]
+  registrationDate: Instant!
+  notes: [ClaimNote!]!
```
meletios, line 948: Do we always have notes? not sure if this is true when the claim is just created for example.

meletios (commented):
Looks good, one question regarding the schema

schema.graphql
vonElfvin: Yea it should always be at least an empty list

palmenhq (dismissed):
Beautiful imo! Also nice with the schema improvements. However what about PR description checkboxes? :D Changelog, tested etc?

palmenhq (approved):

PR #294, opened 2021-04-08
Spring cleaning 2021 🍃

palmenhq:
- Upgrade a lot of packages (including emotion react (hence the big diff as they swiched packages))
- Fix hot reloading (to work with react hooks etc) (!)
- Fix some random console errors

# Jira Issue: []

## What?
-

## Why?
-

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


shared/hedvig-ui/form.tsx
```diff
@@ -140,7 +140,7 @@ const FormTextAreaComponent: React.FC<FormFieldProps> = ({
       name={name}
       rules={rules}
       defaultValue={defaultValue}
-      as={<TextArea {...props} autoHeight />}
```
vonElfvin, line 143: How come we remove autoHeight?

shared/hedvig-ui/themes.ts
```diff
@@ -103,6 +104,7 @@ export const lightUiTheme = createMuiTheme({
     },
   },
   typography: {
+    useNextVariants: true,
```
vonElfvin, line 107: What is this? 😮

src/components/claims/claims-list/components/ClaimListItem.tsx
```diff
@@ -7,14 +7,20 @@ import { withFadeIn } from 'hedvig-ui/animations/fade-in'
...
 import { useNumberMemberGroups } from 'utils/number-member-groups-context'
 
-const MemberIdCell = styled(Table.Cell)<{
+const MemberIdCell = styled<typeof Table.Cell>(
+  ({
+    memberId: _memberId,
+    numberMemberGroups: _numberMemberGroups,
+    ...props
+  }) => <Table.Cell {...props} />,
+)<{
```
vonElfvin, line 23: Interesting! How come we had to do this?

src/components/dashboard/index.tsx
```diff
@@ -110,8 +110,8 @@ export const Dashboard: React.FC<{ auth: any }> = ({ auth }) => {
             const isRecent =
               differenceInCalendarDays(new Date(), change.date) < 3
             return (
-              <FadeIn delay={`${index * 50}ms`}>
-                <ChangeLogItem key={change.change}>
+              <FadeIn delay={`${index * 50}ms`} key={change.change}>
```
vonElfvin, line 113: Kill those logs! 😍

vonElfvin (approved):
YYYYYEEEESSSSS 😍 😍 😍 😍 😍 

One thing, after speaking with IEX I can confirm that we can kill 🔪 the emoji mart, if we want!

shared/hedvig-ui/form.tsx
rasmusguterstam: Because it does not exist 😅

shared/hedvig-ui/themes.ts
rasmusguterstam: I believe it silences some warnings (and is recommended by these warnings)!

src/components/claims/claims-list/components/ClaimListItem.tsx
palmenhq: It seems `Table.Cell` blindly forwards every prop it receives to the DOM, and since `styled` can't know which props will be forwarded to the DOM by `Table.Cell` it'll forward everything (hence everything is forwarded to the DOM). This way we can omit the props from spreading to the DOM while still pass them as properties to `MemberIdCell`

PR #299, opened 2021-04-12
Fix so we properly show UWGL hit and fix hot reload for components withShowNotification

vonElfvin:
# Jira Issue: []

## What?
- Fix UWGL display bug
- Fix so we properly get hot reload in showNotifcation components by setting `pure: false`

## Why?
- Fix stuff

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/src/components/member/tabs/quote-tab/quote-list-item.tsx
+++ b/src/components/member/tabs/quote-tab/quote-list-item.tsx
@@ -1,12 +1,13 @@
+import styled from '@emotion/styled'
 import { Quote } from 'api/generated/graphql'
 import { UpdateQuoteForm } from 'components/member/tabs/quote-tab/update-quote-form'
 import { format, parseISO } from 'date-fns'
 import { Button } from 'hedvig-ui/button'
+import { ThirdLevelHeadline } from 'hedvig-ui/typography'
 import React from 'react'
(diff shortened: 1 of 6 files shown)


src/components/member/tabs/quote-tab/quote-list-item.tsx
```diff
@@ -61,19 +62,20 @@ const QuoteDetails: React.FC<{
...
-            <h3>Important! Member breaches underwriting guidelines</h3>
-            <ul>
-              {quote.breachedUnderwritingGuidelines!.map((guideline) => (
-                <li key={guideline}>{guideline}</li>
-              ))}
-            </ul>
-          </BreachedUnderwritingGuidelines>
-        </DetailWrapper>
-      ))}
+    {(quote.breachedUnderwritingGuidelines?.length || 0) > 0 && (
```
palmenhq, line 65: Honestly i don't even understand why we need the `|| 0` here. `null > 0 === false` and `undefined > 0 === false` so we don't need to default to 0 imo

palmenhq (approved):
good stuff!

AlexSJ1990 (approved):
⭐️

PR #312, opened 2021-04-27
Add click to copy for ssn in claims view, move files tab to the right

vonElfvin:
# Jira Issue: []

## What?
- Add click to copy to ssn in claims view
- Move files more to the right
- Fix so we can create account entries

## Why?
- Ssn is used for FOSS/GSR checkups and the format was wrong "copy wise" but correct readable wise
- Files is less used and since we use CTRL-1 to CTRL-9 as hotkey, let's move the more used ones to the left a bit
- Account entries were broken

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/shared/hedvig-ui/popover.tsx
+++ b/shared/hedvig-ui/popover.tsx
@@ -3,6 +3,7 @@ import styled from '@emotion/styled'
 import React from 'react'
 
 const Contents = styled.div`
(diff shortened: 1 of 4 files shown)


palmenhq (approved):
Lgmt!

PR #313, opened 2021-04-27
Fix Rendered more hooks than during the previous render

vonElfvin:
# Jira Issue: []

How how this not happened before??

## What?
- Move `useState` further up to hopefully not have different amount of hooks during different renders 🤦 

## Why?
- Sometimes it breaks completely when trying to activate a quote

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/src/components/member/tabs/quote-tab/quote-activation.tsx
+++ b/src/components/member/tabs/quote-tab/quote-activation.tsx
@@ -6,7 +6,7 @@ import {
 import { useContracts } from 'graphql/use-contracts'
 import { Button } from 'hedvig-ui/button'
 import { DateTimePicker } from 'hedvig-ui/date-time-picker'
-import React, { useState } from 'react'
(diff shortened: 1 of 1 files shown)


rasmusguterstam (approved):
💀💀

src/components/member/tabs/quote-tab/quote-activation.tsx
```diff
@@ -48,13 +55,10 @@ export const QuoteActivation: React.FC<{
   if (contract.hasPendingAgreement && contract.isTerminated) {
     return <>Cannot active quote for a pending contract that is terminated</>
```
palmenhq, line 56: all returns must go after the hooks, so i think we need to move down these returns or move up the hooks :)

palmenhq (commented):
im a bit late to the party but please fix in another pr :D

PR #314, opened 2021-04-28
Allow showing quotes when no contract, move up hook in activation, improve tab order slightly

vonElfvin:
# Jira Issue: []

## What?
- Allow showing quotes when no contract exists using pickedLocale as fallback
- Move up `useEffect` in quote activation
- Move Files tab to the left and Member tab to the right
- Add pickedLocale fallback flag to member page

## Why?
- Currently we cannot know what quote a member has if they haven't signed for no real reason than us not using pickedLocale as fallback
- Hooks should not change at runtim
- Files tab is used a lot and member tab almost never. Now the most used tabs are wihting Ctrl-1 to Ctrl-6 reach

## Optional screenshots
image

This was not possible before: 
image
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/components/member/tabs/quote-tab/quote-activation.tsx
```diff
@@ -56,10 +70,6 @@ export const QuoteActivation: React.FC<{
     return <>Cannot active quote for a pending contract that is terminated</>
   }
 
-  useEffect(() => {
-    setActiveFrom(getInitialActiveFrom(contract))
-  }, [])
-
   const [addAgreement, addAgreementMutation] = useAddAgreementFromQuote()
```
palmenhq, line 73: I think this also needs to go before the return

vonElfvin: This one is next to all the other hooks. The first return will never change over time so I just put it on the top since there is no point in continuing if we don't have an `originatingProductId`

vonElfvin: With that the number of hooks would never change, but maybe it should still be underneath all the hooks so it's always the full amount of hooks?

palmenhq: I think its better to keep all the hooks in one place and do it "by the book". Maybe we'll change something and miss that this hook is down here. That could cause very unexpected errors.

vonElfvin: But I have moved the hook up? 🤔 This comment is on the code I have moved...

palmenhq: `useAddAgreementFromQuote()` is also a hook no? Or am i missing something?

vonElfvin: Aha now I see what you mean, I thought you commented on `useEffect`. Fixed now 🙏

palmenhq (approved):
Awesome 🐞🔪

PR #315, opened 2021-04-28
Make so we no longer show contractMarketInfo on memberDetails and pickedLocale not editable

vonElfvin:
# Jira Issue: []

## What?
- Fix member details page
- Add language to member info
- Change name of H.OPE. to Hope
- Add Birthdate celebration support to Hope 😄 
- Allow access to debt tab if no contract is present

## Why?
- Because I brök it
- Language info is nice I guess
- Hope is the future
- Fun thing when it happens
- To be able to whitelist members

## Optional screenshots
image

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally

### DIFF
--- a/src/components/member/tabs/DetailsTab.tsx
+++ b/src/components/member/tabs/DetailsTab.tsx
@@ -46,6 +46,7 @@ const DetailsTabComponent: React.FC<WithShowNotification & {
       case 'birthdate':
       case 'signedon':
       case 'status':
+      case 'pickedlocale':
       case 'createdon':
         return true
       default:
@@ -81,6 +82,7 @@ const DetailsTabComponent: React.FC<WithShowNotification & {
   }
 
   delete memberInfoWithoutSsn.__typename
+  delete memberInfoWithoutSsn.contractMarketInfo
 
   return memberInfoWithoutSsn ? (
     <FadeIn>
--- a/src/components/member/tabs/DetailsTab.tsx
+++ b/src/components/member/tabs/DetailsTab.tsx
@@ -55,7 +55,7 @@ const DetailsTabComponent: React.FC<WithShowNotification & {
   }
 
   const handleChange = (field) => (e) => {
-    const editedMemberDetails = editMemberInfoRequest
+    const editedMemberDetails = { ...editMemberInfoRequest }
     editedMemberDetails[field] = e.target.value
     setEditMemberInfoRequest(editedMemberDetails)
   }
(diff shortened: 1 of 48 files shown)


src/components/member/tabs/DetailsTab.tsx
```diff
@@ -81,6 +82,7 @@ const DetailsTabComponent: React.FC<WithShowNotification & {
   }
 
   delete memberInfoWithoutSsn.__typename
+  delete memberInfoWithoutSsn.contractMarketInfo
```
rasmusguterstam, line 86: This feels really weird, why are we doing this? And are we really sure we want to do this outside any `useEffect`? 👀

src/components/member/tabs/debt-tab/index.tsx
```diff
@@ -15,37 +14,52 @@ import { MainHeadline } from 'hedvig-ui/typography'
...
         Issue retrieving debt flag for this member
       </StandaloneMessage>
     )
   }
 
+  const eligibleForWhitelist =
+    !person?.status?.whitelisted &&
+    (person?.debt?.paymentDefaults?.length !== 0 ||
+      (person?.debt?.totalAmountDebt.amount ?? 0) > 0)
+
```
rasmusguterstam, line 62: Maybe make this into a more readable function, with some early returns instead?

src/graphql/use-get-quotes.tsx
```diff
@@ -24,7 +24,7 @@ export const useQuotes = (memberId: string): GetQuotesReturnTuple => {
       memberId,
     },
   })
-  const quotes = queryResult.data?.member?.quotes as Quote[] | undefined
+  const quotes = queryResult.data?.member?.quotes
   const contractMarket =
     queryResult.data?.member?.contractMarketInfo ?? undefined
```
rasmusguterstam, line 29:
```suggestion
  const contractMarket =
    queryResult.data?.member?.contractMarketInfo
```

src/utils/contract.tsx
```diff
@@ -9,7 +9,7 @@ export const getSignSource = (signSource: SignSource): string => {
     case SignSource.Ios:
       return 'iOS'
     case SignSource.Hope:
-      return 'H.OPE.'
```
rasmusguterstam, line 12: nice

rasmusguterstam (approved):
Really nice, and sweet with some tests! Just some minor comments / questions.

image

src/graphql/use-get-quotes.tsx
vonElfvin: It can be null tho 😢

rasmusguterstam: oh

src/components/member/tabs/DetailsTab.tsx
vonElfvin: As of now yes, this should go into the bin at some point though after being done properly with graphql

PR #320, opened 2021-05-18
Fix: Remove double audio stream

rasmusguterstam:
## What?
- Remove the double audio in Claim Information
- Add "Download button" next to audio
- Add a Label to hedvig-ui

## Why?
- Don't know why we try to stream an unsigned s3 link (which seemingly never work), so let's skip that (unless I'm missing something obvious here??)

### Before
image

### After
image

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


vonElfvin (approved):
🙈

PR #326, opened 2021-06-01
Update StandaloneText, add more carrier info to claim

vonElfvin:
# Jira Issue: []

## What?
- Show carrier info more often on claims page
- Do not allow settings reserves if no carrier
- Make so standalone text can handle html elements without being weird
- Add `yarn fix` command

## Why?
- So we don't do payouts incorrectly
- So we don't do reserves incorrectly
- So I don't have to run both commands myself

## Optional screenshots
image

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


shared/hedvig-ui/animations/standalone-message.tsx
```diff
@@ -1,23 +1,21 @@
...
 
 interface StandaloneMessageProps {
+  opacity?: number
   paddingTop?: string
   paddingBottom?: string
   paddingLeft?: string
   paddingRight?: string
 }
 
-export const StandaloneMessage = styled('div')<StandaloneMessageProps>`
```
rasmusguterstam, line 13: Why are we creating an entire `FC` for the `StandaloneMessage`, instead of just extending `StandaloneMessageProps` and let it be a styled component? It doesn't seem as if we're adding anything specific?

src/components/claims/claim-details/components/ClaimPayment.tsx
```diff
@@ -48,16 +47,14 @@ export interface PaymentFormData {
...
-const SubmitButton = withStyles({
-  root: {
-    marginTop: '1rem',
-    display: 'block',
-  },
-})(MuiButton)
+const SubmitButton = styled(MuiButton)`
+  margin-top: 1rem;
+  display: block;
+`
```
rasmusguterstam, line 57: Since it's a Material-UI button, you probably want to do something like:

```suggestion
const SubmitButton = styled(MuiButton)`
  && {
    margin-top: 1rem;
    display: block;  
  }
`
```

It might look good locally, but when you bundle in staging and production there's a risk it overrides the styles (which you prevent with this)

src/components/claims/claim-details/components/ClaimPayments.tsx
```diff
@@ -160,28 +159,14 @@ export const ClaimPayments: React.FC<Props> = ({ claimId }) => {
...
-            </Spacing>
-          )}
-        </>
+        <ClaimPayment
+          sanctionStatus={paymentsData?.claim?.member.sanctionStatus}
+          claimId={claimId}
+          refetch={refetchPayments}
+          identified={Boolean(identity)}
+          market={paymentsData?.claim?.contract?.market!}
+          carrier={paymentsData?.claim?.agreement?.carrier!}
```
rasmusguterstam, line 168: Maybe go for a conditional render instead of null-asserting? 🧐

src/components/claims/claim-details/graphql/claim-page-query.graphql
```diff
@@ -0,0 +1,14 @@
...
+    state
+    contract {
+      id
+    }
+    agreement {
+      id
+      carrier
+    }
+  }
+}
```
rasmusguterstam, line 14: Nice to create a query for a specific view!

src/utils/text.tsx
```diff
@@ -26,3 +26,15 @@ export const formatPostalCode = (postalCode: string): string => {
...
+  switch (carrier) {
+    case 'EIR':
+      return '⚠️ EIR'
+    case 'HEDVIG':
+      return 'Ⓗ Hedvig'
+    case 'HDI':
+      return '🗄 HDI'
+  }
+  return carrier
+}
```
rasmusguterstam, line 40: No need to change this, but instead of doing these switch statements in the future, I think it's more clean to start using `Records`. So in this case, something like:

```typescript

type Carrier = 'HDI' | 'HEDVIG' | 'EIR'

export const carrierDisplayName: Record<Carrier, string> = {
  HEDVIG: 'Ⓗ Hedvig',
  HDI: '🗄 HDI',
  EIR: '⚠️ EIR',
}

```

Obv you don't even need to define a type explicitly, and could instead just go for `Record<string, string> = { ... }`

rasmusguterstam (dismissed):
Nice, just some minor comments!

shared/hedvig-ui/animations/standalone-message.tsx
vonElfvin: Otherwise it'll behave very weirdly if someone wants to add a `<strong>` inside the text for example. Since it will creat a new line break for each

src/utils/text.tsx
vonElfvin: I agree that's nicer, unsure how it'll behave if were to get something different than these ones though? 🤔

rasmusguterstam (approved):

PR #329, opened 2021-06-08
don't allow gaps in insurance

AlexSJ1990:
# Jira Issue: []

## What?
- We will no longer allow members to have gaps between agreements.
- If a member is terminated and wants to rejoin Hedvig after a gap in time, we will need to create a new contract for them. This can be done using the create button in the quotes tab -> from here we can sign the contract.
- If iex try to create a new quote from an agreement of a terminated contract, they will no longer be able to activate with a gap between the last agreement and the new agreement. They will only be able to activate the quote on the day after the termination date of the previous agreement.

## Why?
- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/components/member/tabs/quote-tab/quote-activation.tsx
```diff
@@ -104,54 +98,26 @@ export const QuoteActivation: React.FC<{
...
-                setUseGap(!!checked)
-              }}
-              label="Create gap between insurances"
-              checked={useGap}
-            />
-          </BottomSpacerWrapper>
-
-          {useGap && (
-            <BottomSpacerWrapper>
+            {contract.isTerminated && (
```
vonElfvin, line 102: We will move away from `isTerminated` eventually
```suggestion
            {contract.terminationDate && (
```

src/components/member/tabs/quote-tab/quote-sub-section.tsx
```diff
@@ -20,16 +20,11 @@ export const QuotesSubSection: React.FC<{
...
-  const hasActiveQuotes = activeQuotes.length > 0
-  const hasSignedQuotes = signedQuotes.length > 0
 
   return (
     <Wrapper>
-      {!hasActiveQuotes && !hasSignedQuotes && (
-        <Button onClick={() => setIsWip(!isWip)}>Create</Button>
-      )}
-      {!hasActiveQuotes && !hasSignedQuotes && isWip && (
+      {<Button onClick={() => setIsWip(!isWip)}>Create</Button>}
```
vonElfvin, line 26: We really don't want IEX to create quotes if the contract is active, can we at least check that it is terminated before allowing this button?
```suggestion
      {(!hasSignedQuotes || contract.terminationDate) && <Button onClick={() => setIsWip(!isWip)}>Create</Button>}
```

vonElfvin (commented):
Looks nice! :) A lot simpler than before.

Two questions:
1. I'm guessing we have some validation in the backend that does not allow parallel contracts 
2. How will we communicate this to IEX? Have we talked to them already about this change?

meletios (approved):
👍

PR #334, opened 2021-07-05
get rid of a bunch of errors in back-office by not calling the API illegally

bystam:
# Jira Issue: []

## What?
- Don't fire request if we don't have enough data for it
- It's also weird that back-office treats a case of "illegal input format" as an error and simply just not a 400+warning

## Why?
- See screenshot
image

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


src/components/claims/claim-details/components/claim-items/components/ValuationInfo.tsx
```diff
@@ -21,15 +21,14 @@ export const ValuationInfo: React.FC<{
...
   const { itemFamilyId, itemTypeId, purchasePrice, dateOfPurchase } = request
-
   const [
     valuationStatus,
     { loading: loadingValuation },
   ] = useCanValuateClaimItem(itemFamilyId, itemTypeId, typeOfContract)
 
   const [claimItemValuation] = useGetClaimItemValuation({
-    purchasePrice: purchasePrice ?? 0,
-    itemFamilyId: itemFamilyId ?? null,
```
bystam, line 32: These default values actually hit back-office, which does not accept them

vonElfvin (changes requested):
I think this is not the best way to go about solving this problem, actually I've already given it a stab to solve it on my end when fixing some other stuff in `hedgehog`. If it's fine with you if I ping you there later to "request changes"? 🙏 🤔

bystam:
> I think this is not the best way to go about solving this problem, actually I've already given it a stab to solve it on my end when fixing some other stuff in `hedgehog`. If it's fine with you if I ping you there later to "request changes"? 🙏 🤔

And this isn't a decent first step to try and silence the logs? #tech-on-call is currently quite loud because of this particular problem.

vonElfvin:
> > I think this is not the best way to go about solving this problem, actually I've already given it a stab to solve it on my end when fixing some other stuff in `hedgehog`. If it's fine with you if I ping you there later to "request changes"? 🙏 🤔
> 
> And this isn't a decent first step to try and silence the logs? #tech-on-call is currently quite loud because of this particular problem.

It's more the code in itself that you are changing is not the preferred approach in how we use `React` today. We solve the problem here but it'll be immediately deprecated code wise, partly because everything else is. I have an iteration of the code that both solves the problem and is a small iteration on the code. 

I think we can live a day with some extra logs? Otherwise you are feel to deploy this and I'll just remove it later on in my PR. You decide what you prefer 🙏

bystam:
> > > I think this is not the best way to go about solving this problem, actually I've already given it a stab to solve it on my end when fixing some other stuff in `hedgehog`. If it's fine with you if I ping you there later to "request changes"? 🙏 🤔
> > 
> > 
> > And this isn't a decent first step to try and silence the logs? #tech-on-call is currently quite loud because of this particular problem.
> 
> It's more the code in itself that you are changing is not the preferred approach in how we use `React` today. We solve the problem here but it'll be immediately deprecated code wise, partly because everything else is. I have an iteration of the code that both solves the problem and is a small iteration on the code.
> 
> I think we can live a day with some extra logs? Otherwise you are feel to deploy this and I'll just remove it later on in my PR. You decide what you prefer 🙏

I am 100% fine with either. I was just wondering whether it's going to be like a "don't do this now, we'll fix it in 6 months when we have rewritten all of Hedgehog". Then I think it would make sense to merge this now. But if you are close to fixing it elsewhere this week then that sounds cool :)

bystam:
Closing because it will be fixed in a different PR by Elvin

PR #339, opened 2021-07-09
Ipc 180 add islocked info to contract

joacimastrom:
# Jira Issue: IPC-180

## What?
- Show a user when a contract is locked
- Overlay the contract so it cannot be interacted with
- Make `<Card>` a component that takes a prop `locked`
- Minor cleanup of console errors

## Why?
- There is no user feedback for when a contract is locked
- When a contract is locked, all requests will return errors. This is poor user experience

## Optional screenshots
image

## Optional checklist
- [X] Codescouted
- [ ] Unit tests written
- [X] Tested locally

### DIFF
--- a/shared/hedvig-ui/card.tsx
+++ b/shared/hedvig-ui/card.tsx
@@ -38,7 +38,6 @@ export interface CardProps {
   span?: number
   padding?: PaddingSize
   locked?: boolean
-  to?: string
   children: any
 }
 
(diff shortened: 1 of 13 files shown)


shared/hedvig-ui/card.tsx
```diff
@@ -35,8 +53,23 @@ export const Card = styled('div')<CardProps>`
   background-color: ${({ theme }) => theme.accentLighter ?? colorsV3.white};
   border: 1px solid ${({ theme }) => theme.border ?? colorsV3.gray300};
   border-radius: 0.5rem;
+  position: relative;
 `
 
+export const CardLink = CardContainer.withComponent(Link)
+
+export const Card = ({ children, locked, ...cardProps }: CardProps) => (
```
rasmusguterstam, line 61: Just curious, why not `export const Card: React.FC<CardProps> = ...`?

rasmusguterstam (approved):
Nice one!

shared/hedvig-ui/card.tsx
joacimastrom: Just think this syntax is a bit cleaner :) They do exactly the same thing 🤷‍♂️

joacimastrom: Read up on it a bit after this comment, and there is a very minor differense but 

> In most cases it makes very little difference which syntax is used, but the React.FC syntax is slightly more verbose without providing clear advantage, so precedence was given to the "normal function" syntax.

https://stackoverflow.com/a/59991281

PR #346, opened 2021-07-22
Fix claims not updating when choosing claim type

joacimastrom:
## What?
- Remove a lot of small queries getting claims informations, and replace with a large one
- - Request will be sent once on load, all consecutive usages of the query will instead read from graphql cache
- - Remove refetches since components will update when cache does
- Replaced own implementation of `fullWidth` with existing prop `fluid` in semantic-UI input to avoid remove console error of invalid html attribute
- Renamed  prop `loading` with `isLoading` console error of invalid html attribute

## Why?
- Cache is not being updated correctly when choosing type of claim, forcing user to reload the page to refetch the query.

### Amount of requests when loading claim before:
image

### Amount of requests when loading claim now:
image

## Optional checklist
- [X] Codescouted
- [ ] Unit tests written
- [X] Tested locally // Could not reproduce error locally, but at least nothing breaks from these changes


AlexSJ1990 (approved):
Really nice!!

PR #349, opened 2021-08-09
Fixes: low hanging fruit 

rasmusguterstam:
Fixing problems that are so minor that they never get fixed 

- Make `hedvig-ui/input` have a border when focused
- Make entire chat button clickable, instead of just the chevron icon
- Add `min-width` to popovers, so they don't look so weird whenever used
- Use `hedvig-ui/input` instead of `semantic-ui` directly


anthansson (commented):
Seems fine :-)

PR #355, opened 2021-08-12
Create Copyable component

rasmusguterstam:
## What?
- Add `Copyable` component to `hedvig-ui`
- Add this bad boy to Claim: Member Information as requested

## Why?
- For some reason, IEX could not copy the link 

## Optional screenshots

https://user-images.githubusercontent.com/51323202/129224431-9ca6793d-e309-446f-b958-1467e3a814a1.mov

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


shared/hedvig-ui/copyable.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export const Copyable: React.FC<{
+  children: React.ReactNode
+  onClick: () => void
+  copyLabel?: PopoverLabel
+}> = ({ copyLabel, children, onClick }) => {
+  const [copied, setCopied] = useState(false)
+
+  const { before = 'Copy', after = 'Copied!' } = copyLabel ?? {}
+
```
joacimastrom, line 56: Maybe reset copied state after a short delay? :)
```suggestion
useEffect(() => setTimeout(() => setCopied(false), 4000), [copied])
```

joacimastrom: useEffect also needs to be imported on top

shared/hedvig-ui/popover.tsx
```diff
@@ -66,16 +66,8 @@ export const Popover: React.FC<PopoverProps> = ({
...
-        if (onOpen) {
-          onOpen()
-        }
-      }}
-      onMouseLeave={() => {
-        if (onClose) {
-          onClose()
-        }
-      }}
+      onMouseOver={() => onOpen?.()}
```
joacimastrom, line 69: Wow, didn't know you could do this for function calls

joacimastrom:
While the animation looks very nice (👏 ) I think the UX might be more clear and intuitive if we just always show the copy-icon?  Especially since the link and the icon does different things, would be nice to be able to copy without having to hover first 🤷‍♂️

shared/hedvig-ui/copyable.tsx
rasmusguterstam: Oo that's really neat!

joacimastrom (approved):

PR #360, opened 2021-08-16
Add shortcut for notes and make them optimistic

rasmusguterstam:
## What?
- Create shortcut to add claim notes (option + enter)
- Make `addClaimNote` mutation optimistic, which creates a nice feeling of direct feedback

## Why?
- Because IEX should not have to click a button or have to wait for 1s for a very common and simple action

## Before

https://user-images.githubusercontent.com/51323202/129570804-cd21d40f-b91e-42f2-a197-9eb82ad0df70.mov

## After

https://user-images.githubusercontent.com/51323202/129570819-8c3c41fd-fe9a-4e35-bdd9-8d82af871872.mov

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/shared/hedvig-ui/text-area.tsx
+++ b/shared/hedvig-ui/text-area.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import React from 'react'
-import { TextArea as SemanticTextArea } from 'semantic-ui-react'
+import { TextArea as SemanticTextArea, TextAreaProps } from 'semantic-ui-react'
 
 const TextAreaWrapper = styled.div`
   width: 100%;
@@ -17,13 +17,19 @@ export const TextArea: React.FC<{
   placeholder: string
   value: string | undefined
   onChange: (value: string) => void
-}> = ({ placeholder, value: inputValue, onChange }) => {
+} & Omit<TextAreaProps, 'onChange'>> = ({
+  placeholder,
+  value: inputValue,
+  onChange,
+  ...props
+}) => {
   return (
     <TextAreaWrapper className={'ui form'}>
       <StyledSemanticTextArea
         placeholder={placeholder}
         value={inputValue}
         onChange={(_, { value }) => onChange(value as string)}
+        {...props}
       />
     </TextAreaWrapper>
   )
--- a/src/components/claims/claim-details/components/ClaimNotes.tsx
+++ b/src/components/claims/claim-details/components/ClaimNotes.tsx
@@ -6,7 +6,6 @@ import {
 import { format, parseISO } from 'date-fns'
 import { Spinner } from 'hedvig-ui/sipnner'
 import React, { useState } from 'react'
-import { sleep } from 'utils/sleep'
 
 import styled from '@emotion/styled'
 import { PaperTitle } from 'components/claims/claim-details/components/claim-items/PaperTitle'
@@ -45,7 +44,6 @@ const ClaimNoteFooter = styled(Paragraph)`
 const ClaimNotes: React.FC<{ claimId: string }> = ({ claimId }) => {
   const {
     data: claimNotesData,
-    refetch: refetchClaimNotes,
     loading: loadingClaimNotes,
     error: queryError,
   } = useClaimPageQuery({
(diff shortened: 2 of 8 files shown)


shared/hedvig-ui/text-area.tsx
```diff
@@ -17,13 +17,19 @@ export const TextArea: React.FC<{
...
+  onChange,
+  ...props
+}) => {
   return (
     <TextAreaWrapper className={'ui form'}>
       <StyledSemanticTextArea
         placeholder={placeholder}
         value={inputValue}
         onChange={(_, { value }) => onChange(value as string)}
+        {...props}
```
joacimastrom, line 32: Love this. This should be on all our home-made components to ensure native props still work

src/components/claims/claim-details/components/ClaimNotes.tsx
```diff
@@ -42,19 +51,90 @@ const ClaimNoteFooter = styled(Paragraph)`
...
   const [note, setNote] = useState('')
   const [submitting, setSubmitting] = useState(false)
+  const [textFieldFocused, setTextFieldFocused] = useState(false)
+
+  const { data } = useGetMeQuery()
+
+  const isEnterPressed = useKeyIsPressed(Keys.Enter)
+  const isOptionPressed = useKeyIsPressed(Keys.Option)
+
+  React.useEffect(() => {
```
joacimastrom, line 88: Add useEffect to destructured import on top :)

src/components/claims/claim-details/components/ClaimNotes.tsx
```diff
@@ -42,19 +51,90 @@ const ClaimNoteFooter = styled(Paragraph)`
...
+
   const [addClaimNote] = useClaimAddClaimNoteMutation()
   const [note, setNote] = useState('')
   const [submitting, setSubmitting] = useState(false)
+  const [textFieldFocused, setTextFieldFocused] = useState(false)
+
+  const { data } = useGetMeQuery()
+
+  const isEnterPressed = useKeyIsPressed(Keys.Enter)
+  const isOptionPressed = useKeyIsPressed(Keys.Option)
```
joacimastrom, line 86: This feels a bit hacky, can't we add an `onkeyDown`-eventListener on the TextArea and check 
```
const onKeyDown = (e) => {
  if (e. ctrlKey && e. keyCode === 13) {
      handleSubmit()
  }
}
```

src/components/claims/claim-details/components/ClaimNotes.tsx
```diff
@@ -42,19 +51,90 @@ const ClaimNoteFooter = styled(Paragraph)`
...
+    ) {
+      handleSubmitNote()
+    }
+  }, [isEnterPressed, isOptionPressed])
+
+  const handleSubmitNote = () => {
+    const today = getTodayInUTC()
+
+    setSubmitting(true)
+    addClaimNote({
```
joacimastrom, line 104: Might want to consider moving the optimisticResponse-object to a separate function for readability 🤷‍♂️

joacimastrom (commented):
Great improvement, very good proof of concept of optimisticResponse ⭐ 

Would like to change the ctrl+enter check to use the actual event instead of global listeners, if possible

shared/hedvig-ui/text-area.tsx
rasmusguterstam: I know right! :D Really nice to use `Omit` as well to exclude props that we want to do ourselves!

src/components/claims/claim-details/components/ClaimNotes.tsx
rasmusguterstam: Heh yeah that's true, didn't even think about that one

src/components/claims/claim-details/components/ClaimNotes.tsx
rasmusguterstam: So I tried that, but Typescript seems to hate it 🥴 I guess the functionality at least is isolated in `handleSubmitNote`

joacimastrom (approved):
LGTM ⭐

PR #361, opened 2021-08-18
Use LocalStorage instead of window

rasmusguterstam:
## Why

LocalStorage is saved per `protocol://domain:port/`, as opposed to `window` which is per tab

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -13,13 +13,13 @@ const axiosInstance = axios.create({
 })
 
 export const refreshAccessToken = async () => {
-  if ((window as any).__hvg_refreshingAccessToken) {
+  if (localStorage.getItem('hvg:refreshingAccessToken') === 'true') {
     // bail if we're already refreshing
     return
   }
 
   try {
-    ;(window as any).__hvg_refreshingAccessToken = true
+    localStorage.setItem('hvg:refreshingAccessToken', 'true')
     await axios.post('/login/refresh', null, {
       headers: {
         accept: 'application/json',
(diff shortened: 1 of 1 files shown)


src/api/index.ts
```diff
@@ -13,13 +13,13 @@ const axiosInstance = axios.create({
...
 export const refreshAccessToken = async () => {
-  if ((window as any).__hvg_refreshingAccessToken) {
+  if (localStorage.getItem('hvg:refreshingAccessToken') === 'true') {
     // bail if we're already refreshing
     return
   }
 
   try {
-    ;(window as any).__hvg_refreshingAccessToken = true
+    localStorage.setItem('hvg:refreshingAccessToken', 'true')
```
bystam, line 22: We should probably have an expiry time set for this. Maybe like 10-30s?

https://www.sohamkamani.com/blog/javascript-localstorage-with-ttl-expiry/

bystam (approved):

src/api/index.ts
```diff
@@ -12,14 +12,40 @@ const axiosInstance = axios.create({
...
+    localStorage.removeItem(key)
+    return null
+  }
+
+  return item.value
+}
+
 export const refreshAccessToken = async () => {
-  if ((window as any).__hvg_refreshingAccessToken) {
+  if (getItemWithExpiry('hvg:refreshingAccessToken') === 'true') {
```
vonElfvin, line 42: This will never be true since we remove the token? Or am I a stupid ass 🐴

vonElfvin (commented):
Added a thought

src/api/index.ts
bystam: It should be true if this function is called simultaneously across multiple server calls and/or tabs though?

PR #366, opened 2021-08-30
IP-25: Kill auth redux

rasmusguterstam:
# Jira Issue: [IP-25](https://hedvig.atlassian.net/browse/IP-25)

Linked to this one [https://github.com/HedvigInsurance/back-office/pull/389](https://github.com/HedvigInsurance/back-office/pull/389)

## What?
- Remove auth state from redux (which is the final piece, woho!)
- Remove PrivateRoute
- Extend `me` query with a `Me` type

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


src/routes/PrivateRoute.tsx
```diff
@@ -1,33 +1,22 @@
...
       render={(props) => {
-        if (authState === AuthState.UNKNOWN) {
+        if (loading) {
           return (
             <LoadingMessage paddingTop={'25vh'}>Authenticating</LoadingMessage>
           )
         }
 
-        if (authState === AuthState.UNAUTHENTICATED) {
+        if (error || !data?.me) {
```
palmenhq, line 19: what if there's a network glitch (ie the member lose wifi for a couple of minutes)? should you really log the person out in that case? Maybe you could just check the error type and status code and retry with some kind of backoff

src/utils/auth.ts
```diff
@@ -1,3 +1,26 @@
...
 export const forceLogOut = () => {
   window.location.pathname = '/login/logout'
 }
+
+const getCookie = (name: string) => {
+  return document?.cookie
+    .split('; ')
+    .find((row) => row.startsWith(`${name}=`))
+    ?.split('=')[1]
+}
```
palmenhq, line 12: i think you should use a "proper" cookie parser lib actually, this feels a little shaky to diy

src/utils/auth.ts
```diff
@@ -1,3 +1,26 @@
...
+export const useScopes = () => {
+  const { data, loading, error } = useGetMeQuery()
+
+  if (loading || error) {
+    return null
+  }
+
+  if (data?.me) {
+    const cookie = getCookie('_hvg_scope')
+    console.log(cookie)
```
palmenhq, line 23: 🙈

src/server/auth.ts
```diff
@@ -1,7 +1,7 @@
 import { addDays, addMinutes } from 'date-fns'
+import jwt_decode from 'jwt-decode'
```
palmenhq, line 2: I think you've been in python-land too much lately 🐍

src/utils/auth.ts
```diff
@@ -1,3 +1,26 @@
...
+
+const getCookie = (name: string) => {
+  return document?.cookie
+    .split('; ')
+    .find((row) => row.startsWith(`${name}=`))
+    ?.split('=')[1]
+}
+
+export const useScopes = () => {
+  const { data, loading, error } = useGetMeQuery()
```
palmenhq, line 15: if we use the `getMe` query here anyways can't we just use the scopes from that result or so? I can't see the back-office repo anymore but it should be available in the [gatekeeper response](https://github.com/HedvigInsurance/gatekeeper/blob/91800e9343c20bd102d41af4b112937390b23c98/src/main/kotlin/com/hedvig/gatekeeper/GatekeeperApplication.kt#L186-L192) so it shouldn't be too hard to read from there instead

palmenhq (changes requested):
I happened to notice this and figured i'd give my two cents! AWESOME to see redux is finally going to 🔥 😍 i'm assuming there's another ticket for 🔪 the other parts

hope y'all are alright over there, tell the others i said hi!

src/server/auth.ts
rasmusguterstam: This seems to be the only way for this package... 😩

rasmusguterstam:
> I happened to notice this and figured i'd give my two cents! AWESOME to see redux is finally going to 🔥 😍 i'm assuming there's another ticket for 🔪 the other parts
> 
> hope y'all are alright over there, tell the others i said hi!

@palmenhq What a lovely surprise - we were actually just talking about the PRs during standup when someone said "...oh, Palmen just reviewed one of those" 😄 

Hope (🤪) you're alright and that you're killing it at the new place!! 🔥

rasmusguterstam:
And as always - great feedback, cheers!!

src/server/auth.ts
palmenhq: I just ment the `snake_case` :D

palmenhq:
Haha! all is good here thanks, cheers!

src/server/auth.ts
rasmusguterstam: oh god

src/App.tsx
```diff
@@ -124,35 +124,17 @@ const App: React.FC = () => {
                               return null
                             }}
                           />
-                          <Routes.PrivateRoute
```
joacimastrom, line 127: Don't we still need PrivateRoute but with an implementation that uses the `getMeQuery`?

palmenhq: Or maybe something that hides everything behind a loader (similar to before) or white page before we know they're not authenticated, to avoid a content flash.

joacimastrom: Looked into it and realised ApolloClient checks auth already so maybe PrivateRoute is not necessary after all. WDYT?

joacimastrom (approved):
Looks good! Approved assuming it's tested ☺️

src/App.tsx
palmenhq: Yes from a security perspective i don't think it matters, but we'll still get the content flash if you try to access the system without being authenticated.

schema.graphql
```diff
@@ -7,7 +7,7 @@ type QueryType {
   member(id: ID!): Member
   claim(id: ID!): Claim
   paymentSchedule(status: ChargeStatus!): [SchedulerState]
-  me: String
+  me: Me
```
vonElfvin, line 10: With this we probably need to make a "aligned deployment" with back-office+hedgehog at the same time, not the end of the world but something to be aware of

src/App.tsx
```diff
@@ -124,35 +124,17 @@ const App: React.FC = () => {
...
-                                {...routeProps}
-                                store={store}
-                              />
-                            )}
+                            component={Routes.MembersPageRoute}
                           />
-                          <Routes.PrivateRoute
+                          <Route
                             path="/tools"
                             store={store}
```
vonElfvin, line 139: Guess we can remove store here?

src/components/claims/claim-details/components/ClaimNotes.tsx
```diff
@@ -96,7 +96,7 @@ const ClaimNotes: React.FC<{ claimId: string }> = ({ claimId }) => {
           notes: [
             {
               text: note,
-              handlerReference: data?.me ?? '',
+              handlerReference: data?.me?.email ?? '',
```
vonElfvin, line 99: For future reference I guess this should not be sent from the client, but rather extracted in the mutation inside back-office

src/features/tools/employees/index.tsx
```diff
@@ -14,9 +13,10 @@ export const Row = styled.div`
   align-items: center;
 `
 
-export const EmployeesComponent: React.FC<{ scopes: readonly string[] }> = ({
-  scopes,
-}) => {
+export const Employees: React.FC = () => {
+  const { data } = useGetMeQuery()
+  const scopes = data?.me?.scopes ?? []
```
vonElfvin, line 18: In the future I think we'll have to have scopes in some context for easy access, but this will do the trick for sure! 🥇

vonElfvin (approved):
Wow 😍 I don't think content flash is an issue since it'll:
1. Either be 0 data since all requests require authentication
2. If there is content it'll be cached at most, but arguably that is impossible since if you are logged out you leave the site by design

src/components/claims/claim-details/components/ClaimNotes.tsx
rasmusguterstam: Yeah def agree!

palmenhq (approved):
Good stuff 🔥

PR #375, opened 2021-09-08
Feature - create modal component

maxplt:
# Jira Issue: [IP-30](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-30)

## What?
Create prototype of Modal component for hedvig-ui

## Why?
Instead of system windows.confirm

## Optional checklist
- [ ] Codescouted
- [ ] Tested locally

## Video
https://user-images.githubusercontent.com/89198006/132466468-4fba5be9-ba39-498e-9ade-c8b924094389.mov


vonElfvin (commented):
WOW 😍 This looks very promising, the GIF and storybook usage is 10/10

shared/hedvig-ui/modal.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    document.addEventListener('mousedown', listener)
+    document.addEventListener('touchstart', listener)
+
+    return () => {
+      document.removeEventListener('mousedown', listener)
+      document.removeEventListener('touchstart', listener)
+    }
+  }, [ref, handler])
+}
```
rasmusguterstam, line 24: Woah really sweet!

rasmusguterstam (commented):
Agree with Elvin, this is really really good! 🔥 The next steps would be to make it easy to consume - perhaps using a combination of a top-level component and a `Context` with some nice custom hook? Would be really sweet if it could be consumed anywhere like

```typescript
const Component = () => {
    const { confirm } = useConfirmDialog()

    ...

    const handler = () => {
        confirm("Are you sure")
        .then(() => { /* go */ })
        .catch(() => { /* oh shit stop */ })
    }
}
```

shared/hedvig-ui/modal.tsx
```diff
@@ -0,0 +1,137 @@
...
+  display: 'flex',
+  alignItems:
+    position === 'top'
+      ? 'flex-start'
+      : position === 'bottom'
+      ? 'flex-end'
+      : 'center',
+  justifyContent:
+    side === 'left' ? 'flex-start' : side === 'right' ? 'flex-end' : 'center',
+}))
```
rasmusguterstam, line 47: One very very minor thing: we have decided that we go for css tags with styled components instead of objects, e.g.:

```typescript
const ModalWrapperStyled = styled.div<{
  position?: 'top' | 'center' | 'bottom'
  side?: 'left' | 'center' | 'right'
}>`
width: 100vw;
height: 100vh;
padding: 50px;
background-color: ${({ theme }) => theme.backgroundTransparent};
...
`


```

maxplt: Previously I was using just css tags. But then I checked Button component and use styles like there

rasmusguterstam: Yeah I think that's an old one, it should be with tags as well 😄

shared/hedvig-ui/modal.tsx
```diff
@@ -0,0 +1,144 @@
...
+  padding: 50px;
+`
+
+const ModalContent = styled.div<{
+  width?: string
+  height?: string
+}>`
+  width: ${({ width }) => width || '350px'};
+  max-width: 900px;
+  height: ${({ height }) => height || '350px'};
```
joacimastrom, line 58: Maybe default here should be auto? :)

joacimastrom:
Very nice and thorough core component ⭐  I agree with Rasmus that a couple of hooks for common use cases would be nice but doesn't need to be included in this PR :)

rasmusguterstam (approved):
🔥

PR #376, opened 2021-09-08
Feature: Tables with hedvig-ui

rasmusguterstam:
# Jira Issue: [IP-36](https://hedvig.atlassian.net/browse/IP-36)

This PR was initially set out to simply add a nice generalized `Table` component to `hedvig-ui`, go away from `semantic-ui`, remove some duplicate code and seamlessly integrate keyboard navigation in every table. It turned out to be pretty philosophical w.r.t application structure, ownership of business logic and whether you shouldn't stay DRY in exchange for sanity and scalability.

Some key takeaways are:
  - Generalizing tables to one, huge components will not scale. Tables are by nature fairly complex and vary a lot, and if we keep it as close to native html as possible (and not use a shared and generalized `Paginator` or what not), it will scale better. That's why the `hedvig-ui/table` components are basically just styled `table` `td` `tr` `th` tags.
  - We should go for a feature based approach. This way, we treat features as isolated modules that require little knowledge of the rest of `hedgehog`.
  - Using the hook `useVerticalKeyboardNavigation` explicitly and relating it with the table at hand is probably the best and, in the long run, simplest way to do keyboard navigation. The alternative would be integrating it with the `Table` component out-of-the-box, but considering the fact that tables tend to vary a lot this will probably escalate.
  - We shouldn't be too afraid of repeating ourselves if there's no other way, **as long we use a feature based approach and stay away from coupling components**. If we come up with a new basic component that can be used across a lot of features now and in the future, that's great - we'll add it to `hedvig-ui`. But if the component only generalizes slightly to two other features, it might not be worth the trade-off of coupling. As long as it's easy to deprecate a feature seamlessly and introduce a newer version, we should just get it done.

## What?
- Add `Table` to `hedvig-ui`
- Add `Monetary` to `hedvig-ui/typography` as a simple way to present a slightly formatted monetary amount
- Refactor "Claims List", "Member Claims List" and "Member Search List" to use new Table

https://user-images.githubusercontent.com/51323202/132515310-b89081ba-9756-4da3-9d94-b737408fe276.mov

image

image

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


shared/hedvig-ui/table.tsx
```diff
@@ -0,0 +1,113 @@
...
+  text-align: left;
+  width: 100%;
+  border-collapse: collapse;
+`
+
+export const TableColumn = styled.td`
+  padding: 1.6em 1em;
+  font-size: 1.05em;
+  transition: all 100ms;
+  :hover {
```
joacimastrom, line 19: No need for `:hover` for cursor-attribute, it will implicitly be for hover :)

src/api/index.ts
```diff
@@ -1,9 +1,8 @@
 import axios, { AxiosResponse } from 'axios'
 import { forceLogOut } from '../utils/auth'
-import config from './config'
 
 const axiosInstance = axios.create({
-  baseURL: config.baseUrl,
+  baseURL: '/api/',
```
joacimastrom, line 5: Curious to why this has changed? Does '/api/' proxy to baseUrl somewhere else?

src/components/claims/claims-list/components/LargeClaimsList.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+      if (!claimId || !memberId) {
+        return
+      }
+
+      history.push(`/claims/${claimId}/members/${memberId}`)
+    },
+  })
+
+  if (!claims) {
```
joacimastrom, line 83: your `useListClaims` defaults to returning claims as an empty list, so this will never load :)

Doesn't the graphql-hook return a `loading`-attribute you can use?

joacimastrom (approved):
Skimmed through it and it looks good 👌 

Regarding your discussion on a general table component vs solid building blocks: if we were to create a general table down the line, I think ant-design makes a really good one to take inspiration from. General yet very customizeable: 
https://ant.design/components/table/

src/api/index.ts
rasmusguterstam: The `config.baseUrl` originally came from a redux config-file, and as this is the only place where it's used nowadays I figured we could remove the old file 😄

rasmusguterstam:
> Skimmed through it and it looks good 👌
> 
> Regarding your discussion on a general table component vs solid building blocks: if we were to create a general table down the line, I think ant-design makes a really good one to take inspiration from. General yet very customizeable:
> https://ant.design/components/table/

Ouf that's actually a really clean approach! I like the Ant design system in general, should probably look there more for inspiration  👀

shared/hedvig-ui/table.tsx
```diff
@@ -0,0 +1,111 @@
...
+  :first-child {
+    border-radius: 8px 0 0 0;
+  }
+
+  :last-child {
+    border-radius: 0 8px 0 0;
+  }
+`
+
+export const TableRow = styled.tr<{ active?: boolean }>`
```
vonElfvin, line 38: Can we make the padding a bit lower on the y-axis? The rows are so high now 🙈

src/components/claims/claims-list/components/LargeClaimsList.tsx
```diff
@@ -0,0 +1,173 @@
...
+    { claims, page: currentPage, totalPages },
+    listClaims,
+    { loading },
+  ] = useListClaims()
+
+  useEffect(() => {
+    listClaims({ page: page - 1 ?? 0 })
+  }, [page])
+
+  const [currentKeyboardNavigationStep] = useVerticalKeyboardNavigation({
```
vonElfvin, line 70: Could we include this in TableRow "by default" somehow? I think it's hard to code but it'd just be so sweet 🤔

src/components/claims/claims-list/components/LargeClaimsList.tsx
```diff
@@ -0,0 +1,173 @@
...
+  return (
+    <>
+      <Table>
+        <TableRow>
+          <TableHeader>Member</TableHeader>
+          <TableHeader>Date Registered</TableHeader>
+          <TableHeader>Claim Type</TableHeader>
+          <TableHeader>Claim State</TableHeader>
+          <TableHeader>Claim Reserves</TableHeader>
+        </TableRow>
```
vonElfvin, line 97: If we manage to get `verticalKeyboard` into the Rows "automatically"
```suggestion
        <TableHeader>
          <TableHeaderColumn>Member</TableHeaderColumn>
          <TableHeaderColumn>Date Registered</TableHeaderColumn>
          <TableHeaderColumn>Claim Type</TableHeaderColumn>
          <TableHeaderColumn>Claim State</TableHeaderColumn>
          <TableHeaderColumn>Claim Reserves</TableHeaderColumn>
        </TableRow>
```

vonElfvin (approved):
This is for sure a step in the right direction 🥇  Let's see if we can go beyond later as well!

PR #377, opened 2021-09-08
Feature: @hedvig-ui module

rasmusguterstam:
## What?
- Figured I'd do this to kick-off the feature based structure, but for everyone's sake I'll keep these separeted
- Create alias so that hedvig-ui components are accessible by importing directly from `@hedvig-ui`


joacimastrom (commented):
Looks great! Do you know if there's any difference in performance in having one large file for imports?

joacimastrom (approved):

PR #382, opened 2021-09-10
A Feature Based Structure

rasmusguterstam:
# Jira Issue: [IP-38](https://hedvig.atlassian.net/browse/IP-38)

## What?
- A cool 248-files-changes PR 😊 No but really, to see what's actually going on here I'd strongly recommend looking at the individual commits. Some parts were legit renaming imports for 200 files, so you can probably skip those 😅 Anyway, here's the brief:
- Create a top folder called `pages` which corresponds directly to the routes defined in `routes.tsx` in a similar way as NextJS does it. The point here is to separate the `view` part of hedgehog and the `feature` part. As you might notice, a feature and a view may often appear to be the same thing but just separated; that's fine, but in general the advantage is that you can retrieve top-level information (e.g. route params, run certain queries etc) and then provide directly to a feature.
- Use similar notation as NextJS to indicate that the route expects one or more params: `member/[memberId]/[tab].tsx`
- Move everything in `src` that isn't a page to `features`. This folder aims to simply gather different features and sub-features (can also be thought of as modules) that are as decoupled from other features as possible. The point is that a new developer introduced to hedgehog shouldn't have to know every other part of the codebase and how everything is coupled; if you have a feature, that's your scope.
- It's worth noting that the current directories under `features` not necessarily align perfectly with how a feature should be defined, but for simplicity I put them there for now and just fixed som low hanging fruit.

## Overview of new directory structure
image

## Optional checklist
- [X] Codescouted
- [ ] Unit tests written
- [X] Tested locally


fd-pliska (commented):
This looks fantastic! Great job!
(I'm not competent enough in front-end to approve this PR, but from the architectural standpoint, you have my applause)

rasmusguterstam:
> This looks fantastic! Great job!
> (I'm not competent enough in front-end to approve this PR, but from the architectural standpoint, you have my applause)

Cheers! 😄 I would even say that front-end knowledge is pretty irrelevant here, as it's more a question of architecture and actually very similar to the one usually adopted in back-end projects! We're trying to do more of that and less of the infamous and slightly chaotic "javascript" way 😅

src/features/member/tabs/DetailsTab.tsx
```diff
@@ -21,6 +20,12 @@ const memberFieldFormatters = {
   createdOn: (date) => dateTimeFormatter(date, 'yyyy-MM-dd HH:mm:ss'),
 }
 
+const isClient = typeof window !== 'undefined'
+
+export const WideModal = styled(Modal)({
```
vonElfvin, line 25: 🔪👁️ 
```suggestion
export const WideModal = styled(Modal)`
```

vonElfvin (approved):
Looks really good to me, let's start here and pick up the rest later 🙏 Would love to hear @joacimastrom's and @maxplt's input on this as well!

joacimastrom:
Empty file

image

rasmusguterstam:
> Empty file
> 
> image

Good catch, it was redundant so no dead links at least 😄

maxplt:
It looks perfect!!! 🤩🤩🤩

joacimastrom (commented):
Okay so I'm gonna have to play the devils advocate here, but if you guys all agree this is the way you want to go I say go for it 🚀  these are just my personal preferences. 

Personally, I think a structure that is based on what is presented makes more sense than what is required to render the view. 

For example, if I want to go to the claims page in my IDE, I would want to be able to search for claims. 
image

As you can see, those files are not at all what I want, instead would have to perform this search which assumes I already know that memberId is required for that view:
image

Furthermore, I don't particularly see the benefit in breaking things out into a feature (e.g component) folder when they are just used in one place. That will eventually make that folder quite cluttered, and you wouldn't know if you can safely make changes in a file without it affecting some other page unless you make a project-wide search first. If you instead had the components used on only one page contained in that pages folder, you can change it without risking any side effects.

## PS.
And on a somewhat unrelated note, when looking at the structure of `pages/[claimId]/members/[memberId]`, I thought "hmm that looks strange". Why do we need the memberId in the url? Can't we get the memberId from the claim on the claimPage? The breadcrumbs don't really make sense the way it is now, navigating "up" yields an empty page
image

rasmusguterstam:
> Personally, I think a structure that is based on what is presented makes more sense than what is required to render the view.
> 
> For example, if I want to go to the claims page in my IDE, I would want to be able to search for claims.
> image
> 
> As you can see, those files are not at all what I want, instead would have to perform this search which assumes I already know that memberId is required for that view:
> image


Although I clearly see the issue and get what you mean, I would argue that the previous "structure" we've had (i.e. things sort of spread out everywhere, but without any clear context or specific naming convention) is way more difficult to navigate. Sure, this new approach won't let you navigate to the page by its file name - but what it does is that it provides a guaranteed single point for every route and view. Also, by renaming every page component to make sure it ends with a `XXXPage`, and if this convention is followed (or enforced with some sort of linting/typescript thingy, that would be really cool), it becomes super easy to find whatever page you're looking for.

> Furthermore, I don't particularly see the benefit in breaking things out into a feature (e.g component) folder when they are just used in one place. That will eventually make that folder quite cluttered, and you wouldn't know if you can safely make changes in a file without it affecting some other page unless you make a project-wide search first. If you instead had the components used on only one page contained in that pages folder, you can change it without risking any side effects.

This is a very good point, and I'd say this is the most difficult trade-off. The thing is that the end-goal is sort of the same (i.e. aim for limited domains which aren't too coupled with others), with the slight difference that we would mix everything together. My personal opinion is that I really enjoy the clean `/pages` hierarchy after using it with NextJS, and the fact that it promotes creating features/modules/whatever-you-want-to-call-them which are testable and limited in their scopes. I think the `MemberPage` is a good example of this, as it clearly separates view & logic:

```typescript

export const MemberPage: React.FC<RouteComponentProps<{
  memberId: string
  tab: string
}>> = (props) => {
  const memberId = props.match.params.memberId
  const [member] = useGetMemberInfo(memberId)

  if (!member) {
    return null
  }

  return <MemberTabs member={member} {...props} />
}

```

- The `MemberPage` only cares about retrieving the route params and resolving the `member` as requested by the "feature" `MemberTabs`
- The `MemberTabs` feature is just a component which explicitly requires a member, and then it will do its thing (and some props in this case haha, but you get the point).
- We can easily add more features to the `MemberPage`, but the abstraction remains the same: the `MemberPage` is the view and it is a one-to-one mapping of the corresponding route. One can do whatever one pleases in the `feature` part, but the backbone and only thing that's consistent and sacred would be the `/pages` folder ✝️ 🙏

> And on a somewhat unrelated note, when looking at the structure of pages/[claimId]/members/[memberId], I thought "hmm that looks strange". Why do we need the memberId in the url? Can't we get the memberId from the claim on the claimPage? The breadcrumbs don't really make sense the way it is now, navigating "up" yields an empty page

Yeah this is probably a ticket in itself, I totally agree!

joacimastrom:
> > Furthermore, I don't particularly see the benefit in breaking things out into a feature (e.g component) folder when they are just used in one place. That will eventually make that folder quite cluttered, and you wouldn't know if you can safely make changes in a file without it affecting some other page unless you make a project-wide search first. If you instead had the components used on only one page contained in that pages folder, you can change it without risking any side effects.
> 
> This is a very good point, and I'd say this is the most difficult trade-off. The thing is that the end-goal is sort of the same (i.e. aim for limited domains which aren't too coupled with others), with the slight difference that we would mix everything together. My personal opinion is that I really enjoy the clean `/pages` hierarchy after using it with NextJS, and the fact that it promotes creating features/modules/whatever-you-want-to-call-them which are testable and limited in their scopes. I think the `MemberPage` is a good example of this, as it clearly separates view & logic:
> 
> ```ts
> export const MemberPage: React.FC<RouteComponentProps<{
>   memberId: string
>   tab: string
> }>> = (props) => {
>   const memberId = props.match.params.memberId
>   const [member] = useGetMemberInfo(memberId)
> 
>   if (!member) {
>     return null
>   }
> 
>   return <MemberTabs member={member} {...props} />
> }
> ```
> 
> * The `MemberPage` only cares about retrieving the route params and resolving the `member` as requested by the "feature" `MemberTabs`
> * The `MemberTabs` feature is just a component which explicitly requires a member, and then it will do its thing (and some props in this case haha, but you get the point).
> * We can easily add more features to the `MemberPage`, but the abstraction remains the same: the `MemberPage` is the view and it is a one-to-one mapping of the corresponding route. One can do whatever one pleases in the `feature` part, but the backbone and only thing that's consistent and sacred would be the `/pages` folder ✝️ 🙏

This I 100% agree with, splitting logic and view makes things a lot simpler and testable and less cluttered. But I think we can maintain that structure and still have the components or views reside in the same folder as the page, where the page is the top level component. Would look something like this: 

```
/src
   /pages
      /MembersPage
         MembersPage.tsx
         /MemberTabs
            MemberTabs.tsx
         /OtherView
            OtherView.tsx

         /components
            SomeSharedComponentsForMembers.tsx
```

Again, this is just my personal preference :)

rasmusguterstam:
> This I 100% agree with, splitting logic and view makes things a lot simple, testable and less cluttered. But I think we can maintain that structure and still have the components or views reside in the same folder as the page, where the page is the top level component. Would look something like this:
> 
> ```
> /src
>    /pages
>       /MembersPage
>          MembersPage.tsx
>          /MemberTabs
>             MemberTabs.tsx
>          /OtherView
>             OtherView.tsx
> 
>          /components
>             SomeSharedComponentsForMembers.tsx
> ```
> 
> Again, this is just my personal preference :)

Ok now I get more what you mean, this looks very promising as well. I guess it comes down to preference; @maxplt and @vonElfvin what do you guys think? 😄

vonElfvin:
> > This I 100% agree with, splitting logic and view makes things a lot simple, testable and less cluttered. But I think we can maintain that structure and still have the components or views reside in the same folder as the page, where the page is the top level component. Would look something like this:
> > ```
> > /src
> >    /pages
> >       /MembersPage
> >          MembersPage.tsx
> >          /MemberTabs
> >             MemberTabs.tsx
> >          /OtherView
> >             OtherView.tsx
> > 
> >          /components
> >             SomeSharedComponentsForMembers.tsx
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Again, this is just my personal preference :)
> 
> Ok now I get more what you mean, this looks very promising as well. I guess it comes down to preference; @maxplt and @vonElfvin what do you guys think? 😄

I think both ideas make a lot of sense. I'd say there is a best of both worlds with a pages+feature architecture with more clear file names, i.e. `MemberPage.tsx` over `[memberId].tsx` but still keeping it in the `pages/members` folder

maxplt:
I have no idea what to add. All is said 🙂
For me everything looks pretty good and interesting. I'll be glad to work with it! 😈

PR #385, opened 2021-09-13
Fix: Redeeming campaign codes & set fraudulent status

rasmusguterstam:
# Jira Issue: [IP-46](https://hedvig.atlassian.net/browse/IP-46)

## What?
- Fix bugs with redeeming campaign codes & set fraudulent status. Seems as if the schema hadn't been generated (should we perhaps include this is in the codefresh pipeline in the future? So this doesn't happen?)
- Codescouted campaigns tab a bit to use new components

## Before

image

## After

image

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


maxplt:
Good job! 
It looks much better now 🤩

maxplt (commented):
Good job!
It looks much better now 🤩

maxplt (approved):

PR #387, opened 2021-09-14
Feature/questions tabs

maxplt:
# Jira Issue: [IP-47](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-47)

## What?
Add Tab navigation in Questions page

## Why?
For better keyboard interaction

## Video
https://user-images.githubusercontent.com/89198006/133201380-e5ef6d6d-8c13-439f-88cc-b14439165ddb.mov

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/features/questions/questions-list/QuestionGroupItem.tsx
```diff
@@ -28,19 +45,47 @@ export const QuestionGroupItem: React.FC<QuestionGroupItemProps> = ({
...
 }) => {
   const [isVisible, setVisible] = React.useState(false)
+  const [isFocused, setFocused] = React.useState(false)
+  const [questionFocused, setQuestionFocused] = React.useState(false)
 
   React.useEffect(() => {
     setVisible(true)
   }, [])
 
+  // TODO: Подсказка типа "для фокуса внутрь тыкните Enter, для разфокуса тыкните Esc"
```
rasmusguterstam, line 55: 🇷🇺🇷🇺🇷🇺

maxplt: hahahah

rasmusguterstam (changes requested):
I just tried the branch myself a bit, and while I don't have any specific comments on the implementation in itself (nice done with `tabIndex`!) I believe we can improve the experience.

The first thing I'm thinking is that we don't want the filters to be "tab-able". The trade-off here is that, when IEX comes to this page, they don't want to press tab 20 times before getting to the first question (even though I noticed you can click the question and then tab from there, it doesn't feel intuitive and IEX will unfortunately never get it). What would be intuitive though is that the first tab press selects the first question (and not the filters) and autofocuses on the input field.

When a question is focused, I believe it should imply that the input field is selected and that I can start writing the answer. If I were to press tab again, I should move on to the next question's input field, and so on. We don't want to force the user to press 4 times to tab before even getting to the input field when it's the primary task. If anything, I'd say we should do smarter shortcuts if a question is selected, for example that if `M` is pressed it will go to the affected member or `C` will take you the claim.

I see what you aim to do with the tip "Press Enter to focus inside and Escape to unfocus", but I'd say we can skip this if we autofocus directly on the input field. We don't need to introduce a concept of "focused" and "unfocused" to IEX, as they won't care nor understand 😅 However, I believe we could add this tip below the `TextArea` (as in claim notes) and let the user submit the answer by pressing `Option` and `Enter` (which apparently doesn't work now? It feels like a no-brainer, right? 🧐)

maxplt:
Yeah, you are right
Navigate just between Textareas is best way in this case
Will do it!

maxplt:
Made edits!
https://user-images.githubusercontent.com/89198006/133225451-bc5725ab-90af-4a5f-994b-36071626f9f4.mov

vonElfvin (commented):
Funnily enough I personally enjoyed the first design a bit more 😄 Especially since if we do it that way we could:
1. Use up/down-arrow instead of tab 🤯 
2. Focus the "top" of the Question first for easy reading of the whole question (since it'll be guaranteed to be from the "top"), and after pressing enter you get the input field focused
3. It aligns more with how the long-term vision for the questions tab looks, where we could open the chat instead of having the input field at all 💯 

With that said I'd say:
1. Focus the whole question (as you did first)
2. Allow usage of the up/down arrows in addition to tab (or maybe even remove tab entirely 🤔)

Thoughts @maxplt @rasmusguterstam ?

joacimastrom:
I haven't tested it but could we speed up the animations just a fraction? :) I think it looks really good but in the video it felt just a bit too slow

rasmusguterstam:
> Funnily enough I personally enjoyed the first design a bit more 😄 Especially since if we do it that way we could:
> 
> 1. Use up/down-arrow instead of tab 🤯
> 2. Focus the "top" of the Question first for easy reading of the whole question (since it'll be guaranteed to be from the "top"), and after pressing enter you get the input field focused
> 3. It aligns more with how the long-term vision for the questions tab looks, where we could open the chat instead of having the input field at all 💯
> 
> With that said I'd say:
> 
> 1. Focus the whole question (as you did first)
> 2. Allow usage of the up/down arrows in addition to tab (or maybe even remove tab entirely 🤔)
> 
> Thoughts @maxplt @rasmusguterstam ?

Yeah now that you mention it, arrow keys actually make more sense since we use that in every table as well 😄 And yes it makes sense to focus on the entire question. 

@maxplt I guess the difference compared to the first implementation would be to autofocus on input if enter is pressed (and not anything else in the question), then unfocus if escape is pressed _and_ perhaps even change focused question if an arrow key is pressed and the input field is empty? 🤔 I'm thinking that you don't want to change question if you have a message going on haha. What do you think?

maxplt:
> > Funnily enough I personally enjoyed the first design a bit more 😄 Especially since if we do it that way we could:
> > 
> > 1. Use up/down-arrow instead of tab 🤯
> > 2. Focus the "top" of the Question first for easy reading of the whole question (since it'll be guaranteed to be from the "top"), and after pressing enter you get the input field focused
> > 3. It aligns more with how the long-term vision for the questions tab looks, where we could open the chat instead of having the input field at all 💯
> > 
> > With that said I'd say:
> > 
> > 1. Focus the whole question (as you did first)
> > 2. Allow usage of the up/down arrows in addition to tab (or maybe even remove tab entirely 🤔)
> > 
> > Thoughts @maxplt @rasmusguterstam ?
> 
> Yeah now that you mention it, arrow keys actually make more sense since we use that in every table as well 😄 And yes it makes sense to focus on the entire question.
> 
> @maxplt I guess the difference compared to the first implementation would be to autofocus on input if enter is pressed (and not anything else in the question), then unfocus if escape is pressed _and_ perhaps even change focused question if an arrow key is pressed and the input field is empty? 🤔 I'm thinking that you don't want to change question if you have a message going on haha. What do you think?

I'm not sure. For me navigate just between inputs are more comfortable and fast. 
What if we'll just make shortcuts for "Mark as resolved" and id link?

PR #394, opened 2021-09-17
Feature/claims navigation

maxplt:
# Jira Issue: [IP-48](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-48)

## What?
Add ability to go to different sections in claims view

## Why?
For better usage with keyboard

## Video
https://user-images.githubusercontent.com/89198006/133722470-e286ab15-4ddf-45d1-90a7-2cd8d0981201.mov

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


shared/@hedvig-ui/Dropdown/dropdown.tsx
```diff
@@ -65,16 +74,26 @@ export const EnumDropdown: React.FC<{
...
+      <SemanticDropdown
+        value={value}
+        placeholder={placeholder}
+        options={dropdownOptions}
+        loading={loading}
+        selection
+        fluid
+        onChange={(_, { value: selection }) => onChange(selection)}
+      />
+    </Ref>
```
rasmusguterstam, line 96: Haha I see why we need to do this, but I can't help thinking that we maybe should just start to slowly kill off `semantic-ui` as well if this is how awkward it is (and there's more to it than just this 🥲 ). Not something you should do here obviously, but the components are very easily replacable by just using styled native html elements, which would guarantee both expected HTML behaviour and give us complete control of the design. But that's for another day 🤓

shared/@hedvig-ui/TextArea/text-area.tsx
```diff
@@ -17,20 +22,37 @@ export const TextArea: React.FC<{
...
+
+  useEffect(() => {
+    const setFocus = async () => {
+      await sleep(1)
+      if (textAreaRef.current && focus) {
+        textAreaRef.current.focus()
+      }
+    }
+
+    setFocus()
```
rasmusguterstam, line 43: Perhaps something like this?
```suggestion
    setTimeout( () => {
      if (textAreaRef.current && focus) {
        textAreaRef.current.focus()
      }
    }, 1000)
```

rasmusguterstam: But also - why do we wait one second?

src/pages/claims/ClaimDetailsPage.tsx
```diff
@@ -33,12 +35,40 @@ const ShowEventButtonWrapper = styled.div`
...
 }>> = ({ match }) => {
   const { claimId } = match.params
   const { pushToMemberHistory } = useContext(MemberHistoryContext)
   const [showEvents, setShowEvents] = useState(false)
+  const { registerActions, isHintingOption } = useCommandLine()
+  const [focus, setFocus] = useState(DEFAULT_FOCUSES)
+
+  registerActions(
+    Object.keys(DEFAULT_FOCUSES).map((section, index) => ({
+      label: `Focus on ${section}`,
```
rasmusguterstam, line 65: Won't this one say like `Focus on memberInfo` and `Focus on claimInfo`?

shared/@hedvig-ui/Dropdown/dropdown.tsx
maxplt: @rasmusguterstam 
Yes!!! I agree with you a lot. I have long wanted to ask why do we need `semantic-ui` at all. It's very hard to interact with them using Refs (and we need to use it because of approach to using the keyboard)
In my opinion it will be perfect to remove it. As I think project and components will become more clear.

shared/@hedvig-ui/TextArea/text-area.tsx
maxplt: Because a line break is triggered

src/pages/claims/ClaimDetailsPage.tsx
```diff
@@ -33,12 +35,40 @@ const ShowEventButtonWrapper = styled.div`
   margin-top: 1em;
 `
 
+const ClaimTip = styled.div`
+  position: absolute;
+  right: 1rem;
+  top: 1rem;
```
rasmusguterstam, line 41: Although I like this, just curious of how it works with the current `CardTitle` component? What it does is that it adds a badge to the top right corner in case of error or other warnings - could you verify that it looks alright when these two co-exist? 

Also, what if we style this one to something similar to the vertical menu one, i.e. some sort of shadowed background?

rasmusguterstam (changes requested):
Nice, added some thoughts!

src/pages/claims/ClaimDetailsPage.tsx
maxplt: Yeah, something like this. At the moment I don't know how to make it more clear

shared/@hedvig-ui/Dropdown/dropdown.tsx
rasmusguterstam: Amazing, so glad you think the same! 😄  And yes, especially now when we heavily move towards keyboard navigation this would probably save us a lot of hassle in the long term 😅 Let's talk about it on standup!

src/pages/claims/ClaimDetailsPage.tsx
maxplt: Refactored some styles and check with `CardTitle`. Think it looks pretty well

image

rasmusguterstam: Yes I agree, this works!

rasmusguterstam (approved):
Great job, make sure to resolve conflicts!

PR #395, opened 2021-09-17
Make first version of hotkeys view

maxplt:
# Jira Issue: [IP-53](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-53)

## What?
Redesign view of hotkeys
At the moment only for Side menu. If this view will approved, they will be replaced everywhere

## Why?
For better UI

## Optional screenshots
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


shared/@hedvig-ui/Typography/typography.tsx
```diff
@@ -66,3 +66,24 @@ export const ErrorText = styled.p`
...
+const HotkeyWrapper = styled.div`
+  position: relative;
+`
+
+const HotkeyStyled = styled(Shadowed)`
+  position: absolute;
+  right: -2rem;
+  top: 0;
+  font-size: 13px;
+  background-color: ${({ theme }) => theme.backgroundTransparentContrast};
```
rasmusguterstam, line 79: Really nice! 😄

shared/@hedvig-ui/Typography/typography.tsx
```diff
@@ -66,3 +66,24 @@ export const ErrorText = styled.p`
...
+`
+
+export const HotkeyView = (props) => {
+  return (
+    <HotkeyWrapper>
+      {props.children}
+      {props.isHintingOption && <HotkeyStyled>{props.hotkey}</HotkeyStyled>}
+    </HotkeyWrapper>
+  )
+}
```
rasmusguterstam, line 89: Very very nice! But doesn't Typescript complain about this? 😮

maxplt: Nope 🤔

rasmusguterstam: Also, perhaps the `isHintingOption` should be renamed to something like `hint`, since this component is in `@hedvig-ui` and doesn't care about the implementation?

rasmusguterstam: Because we might go from option to control, and would be cool not to have to change every prop 😄

src/features/navigation/sidebar/VerticalMenu.tsx
```diff
@@ -425,40 +431,55 @@ export const VerticalMenu: React.FC<any & { history: History }> = ({
...
-                      {isHintingOption && latestClaim.current && <Backspace />}
-                    </MenuText>
+                    <HotkeyView hotkey="C" hint={isHintingOption}>
+                      Claims
+                      {/*
+                          For what?
+                          {isHintingOption && latestClaim.current && (
+                            <Backspace />
+                          )}
+                        */}
```
rasmusguterstam, line 441: I think this was/is a way to go back to the latest used claim. Not really sure if it's still active though? Can you check?  In either case we could probably reconsider its position (because I don't see why it should be here)

rasmusguterstam: So: just remove for now and we'll look into this later on

rasmusguterstam (approved):
Really nice! 🌟

src/features/navigation/sidebar/VerticalMenu.tsx
maxplt: Okay!w

maxplt:
And what about view of hotkeys in other places? 
Should I add this new view to them? 
I added it only for Side menu at the moment

rasmusguterstam:
> And what about view of hotkeys in other places?
> Should I add this new view to them?
> I added it only for Side menu at the moment

Since the vertical menu is a bit different, I think we can go with this one first!

src/features/navigation/sidebar/VerticalMenu.tsx
vonElfvin: Yea this is `Option + L` to go to the latest claim, I don't think it's used by anyone atm so we can remove it from here at least

shared/@hedvig-ui/Typography/typography.tsx
```diff
@@ -66,3 +66,24 @@ export const ErrorText = styled.p`
...
+
+const HotkeyStyled = styled(Shadowed)`
+  position: absolute;
+  right: -2rem;
+  top: 0;
+  font-size: 13px;
+  background-color: ${({ theme }) => theme.backgroundTransparentContrast};
+`
+
+export const HotkeyView = (props) => {
```
vonElfvin, line 82: Why not 🤷 Sounds a bit more reusable IMO
```suggestion
export const Hotkey = (props) => {
```

src/features/navigation/sidebar/VerticalMenu.tsx
```diff
@@ -393,7 +393,9 @@ export const VerticalMenu: React.FC<any & { history: History }> = ({
                   }
                 >
                   <House />
-                  <MenuText>Dashborad {isHintingOption && '(D)'}</MenuText>
+                  <HotkeyView hotkey="D" hint={isHintingOption}>
```
vonElfvin, line 396: Maybe even more clear?
```suggestion
                  <HotkeyView hotkey="D" hinting={isHintingOption}>
```

vonElfvin (approved):
Nice! 😍

vonElfvin:
> And what about view of hotkeys in other places?
> Should I add this new view to them?
> I added it only for Side menu at the moment

We could for sure use this elsewhere when we see the opportunity, The `Member` navigation on the claims view comes to mind :) But we can do that later

PR #399, opened 2021-09-20
Feature/claims filters

maxplt:
# Jira Issue: [IP-66](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-66)

## What?
Add Status and Age filter to claims list

## Why?
For convenient interaction and search of claims

## Video
https://user-images.githubusercontent.com/89198006/133960862-157eda6d-24d8-4da7-91eb-fa4a58c669e3.mov

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/graphql/use-list-claims.tsx
```diff
@@ -26,6 +26,9 @@ export const useListClaims = (): ListClaimsReturnTuple => {
           pageSize: options?.pageSize ?? 20,
           sortBy: options?.sortBy ?? 'DATE',
           sortDirection,
+          filterClaimStates: options?.filterClaimStates || [],
```
vonElfvin, line 29: I think this one should be null if empty/not selected

src/pages/claims/list/ClaimsListPage.tsx
```diff
@@ -13,13 +16,26 @@ const ListPage = styled.div`
...
   page?: string
 }>> = ({
   match: {
     params: { page = '1' },
   },
 }) => {
+  const [filters, setFilters] = useInsecurePersistentState<ClaimsFiltersType>(
+    'claims:filters',
+    {
+      filterClaimStates: [],
```
vonElfvin, line 34: null (I think)
```suggestion
      filterClaimStates: null,
```

src/components/claims/filter/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <FilterWrapper>
+      <Dropdown
+        value={filters.filterClaimStates[0] || ''}
+        onChange={changeClaimStateHandler}
+        options={[
+          { key: 0, value: ClaimState.Open, text: 'Open' },
+          { key: 1, value: ClaimState.Closed, text: 'Closed' },
+          { key: 2, value: ClaimState.Reopened, text: 'Reopened' },
+        ]}
+      />
```
vonElfvin, line 49: I don't think a Dropdown is the correct selection tool for this scenario, for example I think showing `Opened` and `Reopened` will be a common use case. Either if we can get the dropdown to select `multi` (which I think should be available) or utilize checkboxes, similar to how we select Member groups/Countries in `Questions` tab.

vonElfvin (changes requested):
Very neat solution I especially like how the filter communicates with the query 💯 

I think the UX for selected the state needs to be revisited, so we can select more granular queries 🙏

src/components/claims/filter/index.tsx
maxplt: @vonElfvin As I know at the moment we can't filter with multiple values. And when it be able I guess we can use dropdown with multi select (as for me it look more clear) or like in Questions tab

rasmusguterstam (commented):
Really nice! 😄 What do you say we add some labels to the fields as well so it's clear how what they filter on?

maxplt:
> Really nice! 😄 What do you say we add some labels to the fields as well so it's clear how what they filter on?

Agree! Will add them!

maxplt:
### Add labels
image

rasmusguterstam:
> ### Add labels
> image

Can we make it so they're aligned to the left? This looks a bit weird 😅

maxplt:
> > ### Add labels
> > image
> 
> Can we make it so they're aligned to the left? This looks a bit weird 😅

Done!

rasmusguterstam (approved):
Great job, just some conflicts to resolve! 😄 🚢

src/components/claims/filter/index.tsx
fd-pliska: They should be, but in the current state they are not - I made a mistake and passing more than one state breaks the query. It's going to be fixed in IP-49, though, so I suggest we wait to deploy filters to production until all three filters are ready. This is on me and I'm terribly sorry.

vonElfvin: No problems, let's assume they work and wait with merging this until they do :)

maxplt:
## Changes
I've maked a little changes to filters 🙂
Add multi select of filters with Checkboxes (Failed to implement an adequate dropdown multi-select)

https://user-images.githubusercontent.com/89198006/135859934-ba951cf8-fcda-440a-a5f1-6ced9514267d.mov

vonElfvin (commented):
Very neat solution both code wise and design wise!

My only suggestion is maybe we can add some country flags, member group colors or maybe emojis like we have it in the Questions filters to make it a bit more 💅 

In addition to already existing stuff. Maybe add something like this to make it a bit more fun 🥳 :
🌊 Complex
📱 Simple

PR #400, opened 2021-09-20
Feature: Conversations 💬

rasmusguterstam:
## What?
- Introduce "Conversations", an experimental alternative to the questions tab 💬
- A first attempt at feature sort-of-A/B-testing
- A stress test of our new feature approach - it is very loosely coupled to the rest of hedgehog
- A cool and simple way to do stepwise animations with promises
- Hopefully a faster way to answer questions, but as it's a first version and purposely implemented as a beta feature, it will require some input and feedback

## Conversations 💬
- Browse between different conversations (in practice different question groups)
- For each conversation, you get general member information and, if applicable, a list of open claims
- If its a member's birthday, it'll let you know - give them a remarkable experience!
- Access to full chat, not just a question group
- Provides an isolated context for a given set of filters
- Keyboard-first (we don't do "send buttons" here 😤) 

## Why?
- Because I wanted to try out some new stuff, and figured that this could be a pretty sweet use-case
- The current questions tab doesn't give that much of a context to what's going on
- If a user doesn't like this flow, they can always switch back to Questions

## Disclaimer
- This feature relies on LocalStorage to do feature flag and onboarding, which is not perfect. However, until Hope get's some sort of user state persistance this should do the trick - especially considering that this is a experimental feature and the main purpose should be to evaluate the usage.

https://user-images.githubusercontent.com/51323202/133965881-5ded693a-0cfd-45a6-896f-ec14160bb776.mov

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


fd-pliska:
Looks incredible!
I think it would benefit from having context clues for switching between Conversations. If a user knows which way to move along the list to get to, say, a conversation with Carl Piehl, it would reduce the time spent finding the desired conversation and provide a more pleasant experience using this feature.

rasmusguterstam:
> Looks incredible!
> I think it would benefit from having context clues for switching between Conversations. If a user knows which way to move along the list to get to, say, a conversation with Carl Piehl, it would reduce the time spent finding the desired conversation and provide a more pleasant experience using this feature.

Very good point, I was thinking something along these lines as well! Will look into some nice solution 😄

shared/@hedvig-ui/Flex/flex.tsx
```diff
@@ -4,11 +4,13 @@ export const Flex = styled.div<{
...
   justify?: string
   align?: string
+  span?: number
   fullWidth?: boolean
 }>`
   display: flex;
   flex-direction: ${({ direction = 'row' }) => direction};
   justify-content: ${({ justify = 'flex-start' }) => justify};
   align-items: ${({ align = 'flex-start' }) => align};
+  flex: ${({ span = 1 }) => span};
```
joacimastrom, line 14: maybe rename this prop to `grow` and set `flex-grow` instead? It is essentially what you are doing except this also has side-effects on `shrink/basis`

shared/@hedvig-ui/animations/fade.tsx
```diff
@@ -0,0 +1,111 @@
+import { keyframes } from '@emotion/react'
```
joacimastrom, line 1: Nice little animation lib :)

src/features/conversations/FilterSelect.tsx
```diff
@@ -0,0 +1,188 @@
...
+} from 'features/conversations/utils/filters'
+import { FilterState, getFilterColor } from 'features/questions/filter'
+import { useQuestionGroups } from 'graphql/use-question-groups'
+import React from 'react'
+import { Shield, ShieldShaded } from 'react-bootstrap-icons'
+import { QuestionGroup } from 'types/generated/graphql'
+import { range } from 'utils/array'
+import { useNumberMemberGroups } from 'utils/number-member-groups-context'
+
+const FilterButton = styled.div<{ selected: boolean }>`
```
joacimastrom, line 16: Any particular reason these are  `styled.div` instead of `styled.button`? Doesn't matter all too much in this case but with buttons you also automatically get a tabIndex and keyListeners. Better for accessibility but also semantics :)

src/features/conversations/FilterSelect.tsx
```diff
@@ -0,0 +1,188 @@
...
+      <ButtonsGroup style={{ justifyContent: 'center', marginTop: '1.0em' }}>
+        <FadeIn
+          delay={`${animationDelay +
+            animationItemDelay * (numberMemberGroups + 1)}ms`}
+        >
+          <FilterButton
+            selected={filters.includes(FilterState.Sweden)}
+            onClick={() => onToggle(FilterState.Sweden)}
+          >
+            Sweden <span style={{ marginLeft: '0.5em' }}>🇸🇪</span>
```
joacimastrom, line 117: Flag icons 🌟  So simple yet effective

src/features/conversations/chat/ConversationChat.tsx
```diff
@@ -0,0 +1,124 @@
...
+`
+
+const ConversationFooter = styled.div`
+  width: 100%;
+  padding: 1em;
+  border-radius: 8px;
+`
+
+const ConversationTextArea = styled(TextArea)`
+  &&&& {
```
joacimastrom, line 26: What does `&&&&` do? Looks hella trippy 😄

src/features/conversations/chat/ConversationChat.tsx
```diff
@@ -0,0 +1,124 @@
...
+              onFocus()
+            }}
+            onBlur={() => {
+              setInputFocused(false)
+              onBlur()
+            }}
+            placeholder={'Your message goes here...'}
+            value={message}
+            onChange={(value) => setMessage(value)}
+            onKeyPress={(e) => {
```
joacimastrom, line 68: Moving this from the render-method would help readability :)

src/features/conversations/member/MemberSummary.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  return (
+    <>
+      <MemberCard
+        onClick={() => {
+          window.open(`/members/${memberId}`)
+        }}
+      >
+        <Flex align={'center'} justify={'space-between'}>
+          <span style={{ fontSize: '1.8em' }}>
```
joacimastrom, line 105: Maybe use `rem`, or does this need to be relative parent?

src/features/conversations/overview/ConversationsOverview.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const [, setOnboarded] = useInsecurePersistentState<boolean>(
+    'conversations:onboarded',
+    false,
+  )
+
+  return (
+    <>
+      <Flex justify={'flex-end'}>
+        <Button
+          size={'small'}
```
joacimastrom, line 26:
```suggestion
          size='small'
```
😏

joacimastrom: Also the line above, and a few other places 😊  Wont comment them all

shared/@hedvig-ui/Flex/flex.tsx
rasmusguterstam: I tried this first, but it didn't really work as expected and after doing some research it seemed as if this was a valid approach (https://www.w3schools.com/cssref/css3_pr_flex.asp) 🧐

src/features/conversations/chat/ConversationChat.tsx
rasmusguterstam: One day we hopefully don't even rely on `semantic-ui` in `@hedvig-ui`, but until that day we need to do these stupid self-reflective selectors to actually change the styles 🥲

rasmusguterstam:
## Updated video

https://user-images.githubusercontent.com/51323202/134041066-12d67701-0641-4ac8-a876-8418a7f5111e.mov

maxplt:
Looks incredible! 🤩 🤩 🤩

maxplt (approved):

shared/@hedvig-ui/Flex/flex.tsx
joacimastrom: It is almost the same, but using it with shorthand and only one value will also set `flex-basis: 0px`, which in most cases is fine. But since we're actually using the css-property `flex`, why not just have the prop be `flex` too instead of `span`? :) Span is used in grids and tables, while flex uses grow/shrink/basis

joacimastrom (approved):
Think it looks really good, very well done and nice UX!

shared/@hedvig-ui/Flex/flex.tsx
rasmusguterstam: Hehe that's a very good point, makes much more sense

PR #412, opened 2021-09-24
Fix imports for storybook

maxplt:
# Jira Issue: [IP-86](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-86)

## What?
As I understand, Storybook do not support some aliases. When I try to start, I get these errors:

image

For now I just set imports as usual with `../../..` And it works
Maybe there is a better solution..

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


rasmusguterstam (changes requested):
As I mentioned during standup, I'd say the better solution for this is to cut off all imports to anything `src` from `@hedvig-ui` and extract the relevant generic utils to `shared/utils` or `@hedvig/utils`!

rasmusguterstam (approved):
Amazing, looks and feels so much better 🤓

PR #413, opened 2021-09-27
Buttons revisited

rasmusguterstam:
## What?
- Just revisit the good 'ol button and its styling
- Make them easier to consume and style
- Introduce:
- `variants: Primary, Secondary, Tertiary`
- `status: Success, Warning, Danger`
- Easy prop to use `react-bootstrap-icons`
- Use `chroma-js` to dynamically add effects based on theme colors

## Optional screenshots

https://user-images.githubusercontent.com/51323202/134859828-a8debd32-e195-4cb2-b568-fff6bfebbfba.mov

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


shared/@hedvig-ui/Button/button.tsx
```diff
@@ -1,185 +1,256 @@
...
+      padding: 0.75rem 1rem;
+    `
+  }
+
+  if (size === 'large') {
+    return css`
+      padding: 1rem 1.5rem;
+    `
+  }
+}
```
maxplt, line 164: Why not like this? 
`
const paddingSizes = {
   small: '0.5rem 0.75rem',
   medium: '0.75rem 1rem',
   large: '1rem 1.5rem'
}
`

And then 
`const padding = ({ size = 'medium' }) => paddingSizes[size]`

rasmusguterstam: Good point, should do that for these simple cases!

PR #421, opened 2021-09-29
Have different modifiers for Windows, fix sending messages for Conversations

vonElfvin:
# Jira Issue: []

## What?
- Have different modifiers for Windows
- fix sending messages for Conversations
- Ignore Enter key if the input is empty

## Why?
- Command does not exist on Windows and the Windows key is not easy to use in Web it seems

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


rasmusguterstam (approved):
let's goo

PR #513, opened 2021-11-17
Feature: Airdrop-ish

rasmusguterstam:
## What?
- Airdop, ish. Let's you share the page you're on directly with anyone else. If they're online, they'll get a notification prompt (a `verbose` notification). If not, it will just be stored as an unread notification for the user.
- The verbose notification is only shown if: Notification is unread, it's marked as verbose, and it was not sent longer than 10s ago. That's to avoid stacking them up if you're offline for a while.

## Why?
- Because it's nice to have a first use of notifications 💬
- And because we should enable IEX to stay in hope and not share stuff with each other in Slack DMs

## Video
(Me sending one to myself - enabled that just to be able to show it, can't be done for real 🤓 )

https://user-images.githubusercontent.com/51323202/142219384-da927911-9519-4d80-a06f-d864bd744031.mov

## Optional checklist
- [x] Changelog updated
- [x] Codescouted
- [x] Tested locally


src/features/navigation/topbar/TopBar.tsx
```diff
@@ -114,11 +123,21 @@ export const TopBar = () => {
 
         <UserMenu />
 
+        <CircleButton
+          onClick={() => {
+            setShowShareModal(true)
+            console.log(showShareModal)
```
vonElfvin, line 129: 🙈

src/features/user/notifications/NotificationsModal.tsx
```diff
@@ -27,12 +27,12 @@ export const NotificationsModal: React.FC<{
...
   const { me } = useMe()
-  const [markNotificationsAsRead] = useMarkNotificationsAsReadMutation()
+  const [markAllNotificationsAsRead] = useMarkAllNotificationsAsReadMutation()
   const history = useHistory()
 
   useEffect(() => {
-    markNotificationsAsRead({
-      optimisticResponse: { markNotificationsAsRead: true },
+    markAllNotificationsAsRead({
+      optimisticResponse: { markAllNotificationsAsRead: true },
```
vonElfvin, line 35: Does this optimistic response even do anything? Shouldn't the return type be notifications and they'll have "read = true"

src/features/user/notifications/graphql/mark-notification-as-read.graphql
```diff
@@ -0,0 +1,3 @@
+mutation MarkNotificationAsRead($notificationId: ID!) {
+  markNotificationAsRead(notificationId: $notificationId)
```
vonElfvin, line 2: For cache update?
```suggestion
  markNotificationAsRead(notificationId: $notificationId): [Notifications!]! {
    id
    isRead
  }
```

src/features/user/share/ShareModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [sharedWith, setSharedWith] = useState<string[]>([])
+
+  const [sharePath] = useSharePathMutation()
+
+  const location = useLocation()
+
+  useKeyIsPressed(Keys.Escape, onClose)
+
+  const users =
+    data?.users?.filter((user) => {
```
vonElfvin, line 77: Subjective 🤷 
```suggestion
  const users = data?.users?.filter((user) => {
```

src/features/user/share/ShareModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+      if (user.email === me.email) {
+        return false
+      }
+
+      return !filter
+        ? true
+        : user.fullName.toLowerCase().includes(filter.toLowerCase())
```
vonElfvin, line 84: For readability
```suggestion
      
      if (user.email === me.email) {
        return false
      }
      
      if (!filter) {
        return true
      }

      return user.fullName.toLowerCase().includes(filter.toLowerCase())
```

src/features/user/share/ShareModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+    }) ?? []
+
+  const handleShare = (user: Omit<User, 'notifications' | 'signature'>) => {
+    setSharedWith((prev) => [...prev, user.id])
+    toast.promise(
+      sharePath({ variables: { path: location.pathname, userId: user.id } }),
+      {
+        loading: 'Sharing page',
+        success: 'Page shared',
```
vonElfvin, line 93:
```suggestion
        success: `Page shared with ${user.firstName}`,
```

src/features/user/share/components/VerboseNotification.tsx
```diff
@@ -0,0 +1,140 @@
...
+  signature: string
+}> = ({ toastId, notificationId, path, message, signature }) => {
+  const [markNotificationAsRead] = useMarkNotificationAsReadMutation()
+
+  const readNotification = () => {
+    markNotificationAsRead({
+      variables: { notificationId },
+      optimisticResponse: {
+        markNotificationAsRead: true,
+      },
```
vonElfvin, line 67: Similar to above I guess this doesn't do anything?

vonElfvin: Maybe the mutation should return Me?

vonElfvin (commented):
Nice! Would be super cool if we could plug this into the CLI as well in the future 🤔 So you don't have to press that circle and select people etc.

src/features/navigation/topbar/TopBar.tsx
rasmusguterstam: Lol

src/features/user/notifications/NotificationsModal.tsx
rasmusguterstam: The optimistic response is required to start writing directly to the cache, which is the `update` method. If we don't specify the optimistic response, we will still wait for the actual response and _then_ write to the cache.

But yes, that's actually a good point, they probably should have notifications as return types! 🤝

src/features/user/share/ShareModal.tsx
rasmusguterstam: I just let prettier do it's thing 🤓

src/features/user/share/ShareModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+const Container = styled.div`
+  padding: 1rem;
+`
+
+// noinspection CssInvalidPropertyValue
+const UserList = styled.div`
+  margin-top: 1rem;
+  max-height: calc(500px - 10.2rem);
+  overflow-y: scroll;
+  overflow: overlay;
```
joacimastrom, line 27: Had to look this up but that's about f'ing time! Stupid scrollbars taking up space

src/features/user/share/ShareModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  ::-webkit-scrollbar-track {
+    border-radius: 8px;
```
joacimastrom, line 30: 🚨🚨 PX 🚨🚨

src/features/user/share/ShareModal.tsx
rasmusguterstam: I know right 😩

src/features/user/share/ShareModal.tsx
rasmusguterstam: Pixel patrol!!! 🚨

src/features/user/share/ShareModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+  useUsersQuery,
+} from 'types/generated/graphql'
+
+const Container = styled.div`
+  padding: 1rem;
+`
+
+// noinspection CssInvalidPropertyValue
+const UserList = styled.div`
+  margin-top: 1rem;
```
joacimastrom, line 24: In a previous project, we have helper classes for margins which was really nice, so you just slapped on a className instead of a style for these things. It's easier to see the spacing directly in the rendering

For example:
`mt-1` was `margin-top: 1rem`, `mt-2` = `margin-top: 2rem`, `mt-half` for 0.5 rem and so on... Also for left/right/bottom, as well as X and Y for top & bottom or right & left

vonElfvin: That's `tailwind` 😄

joacimastrom: > That's tailwind 😄
Nice! Didn't know it was a legit framework, those classes were just defined with some CSS-functions in the project I was in 😄

PR #518, opened 2021-11-19
Feature: Ability to share path using CLI

maxplt:
# Jira Issue: [IP-184](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-184)

## What?
Add ability to share path with Command Line.
Set `'/'` in CLI as **special** symbol. When you type it you can see _"Advanced Commands"_. 
I think in future we can add more commands like that. And it will be more convenient to select them.

## Why?
For fastest sharing.

## Video
https://user-images.githubusercontent.com/89198006/142605667-ef56166e-59c8-4486-9ca6-fc02e0c6a38c.mov

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

### DIFF
--- a/src/features/commands/command-line-hook.tsx
+++ b/src/features/commands/command-line-hook.tsx
@@ -5,6 +5,7 @@ import {
   Keys,
   useKeyIsPressed,
 } from '@hedvig-ui/hooks/keyboard/use-key-is-pressed'
+import { useMe } from 'features/user/hooks/use-me'
 import React, {
   createContext,
   useContext,
@@ -123,6 +124,9 @@ export const CommandLineComponent: React.FC<{
   const [selectedActionIndex, setSelectedActionIndex] = useState(0)
   const [firstActionIndex, setFirstActionIndex] = useState(0)
 
+  const {
+    me: { email: myEmail },
+  } = useMe()
   const { data } = useUsersQuery()
   const [sharePath] = useSharePathMutation()
 
@@ -146,9 +150,13 @@ export const CommandLineComponent: React.FC<{
 
   const setUsersAsResult = () => {
     const name = searchValue.split('@')[searchValue.split('@').length - 1]
-    const users = data?.users?.filter((user) =>
-      user.fullName.toLowerCase().includes(name.toLowerCase()),
-    )
+    const users =
+      data?.users?.filter(
+        (user) =>
+          user.email !== myEmail &&
+          user.fullName.toLowerCase().includes(name.toLowerCase()),
+      ) ?? []
+
     setSearchResult(
       users?.map((user) => ({
         label: user.fullName,
--- a/src/features/commands/command-line-hook.tsx
+++ b/src/features/commands/command-line-hook.tsx
@@ -12,6 +12,12 @@ import React, {
   useRef,
   useState,
 } from 'react'
+import { toast } from 'react-hot-toast'
+import {
+  User,
+  useSharePathMutation,
+  useUsersQuery,
+} from 'types/generated/graphql'
 
 const CommandLineWindow = styled.div`
   position: absolute;
@@ -83,13 +89,14 @@ const ResultItem: React.FC<{
     >
       <FourthLevelHeadline>{action.label}</FourthLevelHeadline>
       <ResultItemContent>
-        {action.keys.map(({ hint }) => (
-          <CharacterBadge key={hint}>
-            <Paragraph style={{ fontSize: '0.8em', fontWeight: 'bold' }}>
-              {hint}
-            </Paragraph>
-          </CharacterBadge>
-        ))}
+        {action.keys &&
+          action.keys.map(({ hint }) => (
+            <CharacterBadge key={hint}>
+              <Paragraph style={{ fontSize: '0.8em', fontWeight: 'bold' }}>
+                {hint}
+              </Paragraph>
+            </CharacterBadge>
+          ))}
       </ResultItemContent>
     </ResultItemWrapper>
   )
@@ -97,7 +104,7 @@ const ResultItem: React.FC<{
 
 export interface CommandLineAction {
   label: string
-  keys: Key[]
+  keys?: Key[]
   onResolve: () => void
 }
 
@@ -116,12 +123,55 @@ export const CommandLineComponent: React.FC<{
   const [selectedActionIndex, setSelectedActionIndex] = useState(0)
   const [firstActionIndex, setFirstActionIndex] = useState(0)
 
-  useEffect(() => {
-    setSearchResult(
-      actions.filter((item) => {
-        return item.label.toLowerCase().includes(searchValue.toLowerCase())
-      }),
+  const { data } = useUsersQuery()
+  const [sharePath] = useSharePathMutation()
+
+  const handleShare = (user: Omit<User, 'notifications' | 'signature'>) => {
+    toast.promise(
+      sharePath({ variables: { path: location.pathname, userId: user.id } }),
+      {
+        loading: 'Sharing page',
+        success: `Page shared with ${user.fullName.split(' ')[0]}`,
+        error: 'Could not share page',
+      },
     )
+  }
+
+  const advancedActions: CommandLineAction[] = [
+    {
+      label: 'Share path',
+      onResolve: () => setSearchValue('/share @'),
+    },
+  ]
+
+  const setUsersAsResult = () => {
+    const name = searchValue.split('@')[searchValue.split('@').length - 1]
+    const users = data?.users?.filter((user) =>
+      user.fullName.toLowerCase().includes(name.toLowerCase()),
+    )
+    setSearchResult(
+      users?.map((user) => ({
+        label: user.fullName,
+        onResolve: () => {
+          handleShare(user)
+          setSearchValue(`/share @${user.fullName}`)
+        },
+      })) || [],
+    )
+  }
+
+  useEffect(() => {
+    if (searchValue === '/') {
+      setSearchResult(advancedActions)
+    } else if (searchValue.includes('/share') && searchValue.includes('@')) {
+      setUsersAsResult()
+    } else {
+      setSearchResult(
+        actions.filter((item) => {
+          return item.label.toLowerCase().includes(searchValue.toLowerCase())
+        }),
+      )
+    }
     setFirstActionIndex(0)
     setSelectedActionIndex(0)
   }, [searchValue])
@@ -171,7 +221,9 @@ export const CommandLineComponent: React.FC<{
     if (searchResult.length === 0) {
       return
     }
-    hide()
+    if (!searchValue.includes('/')) {
+      hide()
+    }
     searchResult[selectedActionIndex].onResolve()
   }, [isEnterPressed])
 
@@ -302,8 +354,8 @@ export const CommandLineProvider: React.FC = ({ children }) => {
   }
 
   const updateActionKeyCodes = () => {
-    actionKeyCodes.current = actions.current.map((action) =>
-      action.keys.map((key) => key.code),
+    actionKeyCodes.current = actions.current.map(
+      (action) => (action.keys && action.keys.map((key) => key.code)) || [],
     )
   }
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -139,10 +139,10 @@ const App: React.FC = () => {
           <MemberHistoryProvider>
             <NumberMemberGroupsProvider>
               <Router history={history}>
-                <CommandLineProvider>
-                  <ConfirmDialogProvider>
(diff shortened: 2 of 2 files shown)


rasmusguterstam (approved):
This is really cool 🤩 

I tried it out myself, and some thoughts:

- Should not care about lower/upper-case
- Maybe we don’t need the slash?
- Would be nice to have some indication that you’re currently writing a command
- When you press enter, the command line should close instead of replacing (you shouldn’t be able to press multiple times on the same)

Thoughts?

src/features/commands/command-line-hook.tsx
```diff
@@ -117,12 +125,62 @@ export const CommandLineComponent: React.FC<{
...
+
     setSearchResult(
-      actions.filter((item) => {
-        return item.label.toLowerCase().includes(searchValue.toLowerCase())
-      }),
+      users?.map((user) => ({
+        label: user.fullName,
+        onResolve: () => {
+          handleShare(user)
+          setSearchValue(`/share @${user.fullName}`)
```
vonElfvin, line 166: I guess hiding is better?
```suggestion
          hide()
```

src/features/commands/command-line-hook.tsx
```diff
@@ -117,12 +125,62 @@ export const CommandLineComponent: React.FC<{
...
+      })) || [],
     )
+  }
+
+  useEffect(() => {
+    if (searchValue === '/') {
+      setSearchResult(advancedActions)
+    } else if (searchValue.includes('/share') && searchValue.includes('@')) {
+      setUsersAsResult()
+    } else {
```
vonElfvin, line 177: This will provide an "index out of bounds" error but the basic idea is there. This way you will get search help if you write "/sh"
```suggestion
    if (searchValue.includes('/share') && searchValue.includes('@')) {
      setSearchResult(advancedActions)
    } else if (searchValue[0] === '/') {
      setUsersAsResult()
    } else {
```

src/features/commands/command-line-hook.tsx
```diff
@@ -117,12 +125,62 @@ export const CommandLineComponent: React.FC<{
...
   const [firstActionIndex, setFirstActionIndex] = useState(0)
 
-  useEffect(() => {
+  const {
+    me: { email: myEmail },
+  } = useMe()
+  const { data } = useUsersQuery()
+  const [sharePath] = useSharePathMutation()
+
+  const handleShare = (user: Omit<User, 'notifications' | 'signature'>) => {
```
vonElfvin, line 134: This is super nice for sure, but it would be worth considering if we can extract this into some separate component 🤔 So we can add more `advancedActions` in an easy way in the future. 

This will be perfect for now though!

vonElfvin (commented):
The speed 🚤 

Added some minor comments to consider :)

rasmusguterstam (approved):

rasmusguterstam (commented):
By the way, make sure you update the `changelog.ts` - they will never know this exist without announcing it!

rasmusguterstam (commented):
Just noticed that it's always visible, perhaps it should filter?

image

maxplt:
## Add new features

https://user-images.githubusercontent.com/89198006/143192478-e43dd915-4d1f-4864-bada-30dfe86593e1.mov

src/features/commands/command-line-hook.tsx
```diff
@@ -84,29 +85,30 @@ const ResultItem: React.FC<{
...
+  keys?: Key[]
   onResolve: () => void
 }
 
 export const CommandLineComponent: React.FC<{
   hide: () => void
   actions: CommandLineAction[]
 }> = ({ hide, actions }) => {
-  const [searchValue, setSearchValue] = useState('')
+  const [searchValue, setSearchValue] = useState<string>('')
```
rasmusguterstam, line 111: This is one of those cases where you actually don't have to specify a type, since it gets the implied type from the default! 😄

src/features/commands/command-line-hook.tsx
```diff
@@ -117,15 +119,26 @@ export const CommandLineComponent: React.FC<{
   const [selectedActionIndex, setSelectedActionIndex] = useState(0)
   const [firstActionIndex, setFirstActionIndex] = useState(0)
 
+  const { advanced, options } = useAdvancedActions(
+    searchValue,
+    setSearchValue,
+    hide,
+  )
```
rasmusguterstam, line 126: What does `advanced` mean? Perhaps rename it so it's easier to use

src/features/commands/use-advanced-actions.ts
```diff
@@ -0,0 +1,203 @@
...
+const getOnlineUsers = (searchValue, users, myEmail) => {
+  const name = searchValue.split('@')[searchValue.split('@').length - 1]
+  const now = new Date()
+
+  const uniqueUsers =
+    users?.filter(
+      (user) =>
+        user.email !== myEmail &&
+        user.fullName.toLowerCase().includes(name.toLowerCase()),
+    ) ?? []
```
rasmusguterstam, line 49:
```suggestion
  const uniqueUsers = getUniqueUsers(searchValue, users, myEmail)
```

src/features/commands/use-advanced-actions.ts
```diff
@@ -0,0 +1,203 @@
...
+    sharePath({ variables: { path: location.pathname, userId: user.id } }),
+    {
+      loading: 'Sharing page',
+      success: `Page shared with ${user.fullName.split(' ')[0]}`,
+      error: 'Could not share page',
+    },
+  )
+}
+
+const getUniqueUsers = (searchValue, users, myEmail) => {
```
rasmusguterstam, line 28: How can there be duplicates? 🤔

src/features/commands/command-line-hook.tsx
```diff
@@ -117,15 +119,26 @@ export const CommandLineComponent: React.FC<{
...
-        return item.label.toLowerCase().includes(searchValue.toLowerCase())
-      }),
-    )
+    if (advanced) {
+      setSearchResult(options)
+    } else {
+      setSearchResult(
+        actions.filter((item) => {
+          return item.label.toLowerCase().includes(searchValue.toLowerCase())
+        }),
```
rasmusguterstam, line 135:
```suggestion
        actions.filter((item) => item.label.toLowerCase().includes(searchValue.toLowerCase())),
```

src/features/commands/command-line-hook.tsx
```diff
@@ -117,15 +119,26 @@ export const CommandLineComponent: React.FC<{
   const [selectedActionIndex, setSelectedActionIndex] = useState(0)
   const [firstActionIndex, setFirstActionIndex] = useState(0)
 
+  const { advanced, options } = useAdvancedActions(
+    searchValue,
+    setSearchValue,
```
rasmusguterstam, line 124: What if we, instead of passing the explicit handler, provide a callback from the hook which signals when it wants to change the search value? So perhaps something like

```typescript

const { advanced, options } = useAdvancedActions(searchValue, hide, (value) => { ... }) 

```

Not sure about the exact way to go, but I'm just not really a fan of passing state handlers directly as props. Usually, it makes sense to decouple slightly and use callbacks instead

src/features/commands/use-advanced-actions.ts
```diff
@@ -0,0 +1,203 @@
...
+    )
+    .sort((u1, u2) =>
+      u1.fullName.toLowerCase() > u2.fullName.toLowerCase() ? 1 : -1,
+    )
+}
+
+export const useAdvancedActions = (
+  searchValue: string,
+  setSearchValue: (value: string) => void,
+  hide: () => void,
```
rasmusguterstam, line 65:
```suggestion
  searchValue: string,
  onChange: (value: string) => void, // or something like this
  onHide: () => void,
```

src/features/commands/use-advanced-actions.ts
```diff
@@ -0,0 +1,203 @@
...
+      setAdvanced(true)
+      setOptions(advancedActions)
+    } else {
+      setOptions([])
+    }
+
+    if (!searchValue.includes('/')) {
+      setAdvanced(false)
+    }
+  }, [searchValue])
```
rasmusguterstam, line 174: This doesn't feel like it scales very well, could we modularize this in some way? Create a file for each advanced action or something, and "register" them in this hook instead?

src/features/commands/use-advanced-actions.ts
```diff
@@ -0,0 +1,203 @@
...
+    {
+      label: 'Search Members',
+      onResolve: () => {
+        setSearchValue('/search @')
+        setOptions([])
+      },
+    },
+  ]
+
+  const setValuesAsOptions = (values, action, field) => {
```
rasmusguterstam, line 109: Types

rasmusguterstam (changes requested):
Love the new advanced options 🌟 Added some feedback!

src/features/commands/use-advanced-actions.ts
```diff
@@ -0,0 +1,203 @@
...
+      label: 'Go to User',
+      onResolve: () => {
+        setSearchValue('/goto @')
+        setOptions([])
+      },
+    },
+    {
+      label: 'Search Members',
+      onResolve: () => {
+        setSearchValue('/search @')
```
vonElfvin, line 103: In my head, the `@` is for other users, whereas in this case you can search for anything really
```suggestion
        setSearchValue('/search ')
```

src/features/commands/use-advanced-actions.ts
```diff
@@ -0,0 +1,203 @@
...
+    } else if (searchValue === '/') {
+      setAdvanced(true)
+      setOptions(advancedActions)
+    } else {
+      setOptions([])
+    }
+
+    if (!searchValue.includes('/')) {
+      setAdvanced(false)
+    }
```
vonElfvin, line 173: Feels like these if statements can be simplified with early returns and maybe a switch statement? 🤷

vonElfvin (commented):
Very darn cool! 😍 

Added some comments :)

Also, I think it'd be a good idea to start off with the share functionality and have the others in a different PR, but let's not fix that now :)

src/features/commands/use-advanced-actions.ts
maxplt: @vonElfvin 
It seems little tricky to track searching value without `@` 😓
Maybe instead of `@` use something like `:` or else 🤔

src/changelog.ts
```diff
@@ -11,6 +11,11 @@ export interface Change {
 }
 
 export const changelog: ReadonlyArray<Change> = [
+  {
+    date: '2021-11-25',
+    change: "Add advanced features in Command Line. Just open it and type '/'!",
```
rasmusguterstam, line 16: "Search for members wherever you are - press option + space and type '/'!"

rasmusguterstam: They don't know what "advanced features" are, and barely what the "Command Line" is 😢

src/features/commands/command-line-hook.tsx
```diff
@@ -117,12 +119,22 @@ export const CommandLineComponent: React.FC<{
   const [selectedActionIndex, setSelectedActionIndex] = useState(0)
   const [firstActionIndex, setFirstActionIndex] = useState(0)
 
+  useAdvancedActions(
+    searchValue,
+    (value: string) => setSearchValue(value),
+    (value: CommandLineAction[]) => setSearchResult(value),
```
rasmusguterstam, line 125: Don't have to specify types here, they are already defined internally (and your IDE will hint them anyway)

```suggestion
    (value) => setSearchValue(value),
    (value) => setSearchResult(value),
```

rasmusguterstam: And also, what do you think about my previous comment on making the input an interface instead of 4 parameters?

src/features/commands/command-line-hook.tsx
```diff
@@ -117,12 +119,22 @@ export const CommandLineComponent: React.FC<{
...
+    hide,
```
rasmusguterstam, line 126:
```suggestion
    () => hide(),
```

src/features/commands/use-advanced-actions.ts
```diff
@@ -0,0 +1,185 @@
+import { differenceInSeconds, parseISO } from 'date-fns'
```
rasmusguterstam, line 1: I'd suggest moving this file to `commands/hooks/use-advanced-actions.ts`, since it's only used internally

rasmusguterstam (approved):
Tried it out locally, and it's really crisp 🌟 🌟 

Pushed a small refactoring of structure. Have a look at the comments and merge whenever you feel ready!

PR #519, opened 2021-11-22
Feature: Improve some styles

maxplt:
# Jira Issue: []

## What?
Make some improvements for styles:
* Remove scroll from all page but add for content part
* Minor restyle of share modal
* And other small stuff

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

### DIFF
--- a/src/features/user/share/ShareModal.tsx
+++ b/src/features/user/share/ShareModal.tsx
@@ -16,13 +16,9 @@ import {
   useUsersQuery,
 } from 'types/generated/graphql'
 
-const Container = styled.div`
-  padding: 1rem;
-`
-
 // noinspection CssInvalidPropertyValue
 const UserList = styled.div`
-  margin-top: 1rem;
+  padding-right: 1.5rem;
   max-height: calc(500px - 10.2rem);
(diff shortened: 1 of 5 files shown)


src/features/user/share/ShareModal.tsx
```diff
@@ -16,13 +16,9 @@ import {
...
-  padding: 1rem;
-`
-
 // noinspection CssInvalidPropertyValue
 const UserList = styled.div`
-  margin-top: 1rem;
+  padding-right: 1.5rem;
   max-height: calc(500px - 10.2rem);
   overflow-y: scroll;
   overflow: overlay;
```
rasmusguterstam, line 24: I don't think we need to do `overflow: overlay` if we're just adding right padding anyway?

rasmusguterstam (approved):
This is awesome, such a no-brainer that only the content should be scrollable 😄 

Noticed that the top-nav doesn't throw a shadow on the content though (only on the background), perhaps it's easy to solve with some z-index magic? Or maybe it is but I can't tell the difference? 😳 

image

PR #548, opened 2021-12-30
Feature: Improve navigation

maxplt:
# Jira Issue: [IP-198](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-198)

### New Features:
Add field `autoFocus` to `useNavigation` hook. So, if on the current page we have a component with `autoFocus`, it'll be focused automatically if current `cursor` is `null`

Implement a new `useNavigation` concept to **main pages** and **components**:
* **Dashborad**
   * Metrics 
* **Claims List Page**
   * Filters
   * Templates  
   * Claims List
* **Member Search Page**
   * Search Form
   * Suggestions
   * Members List 
* **Conversations Page**
   * Filters 
* **Tabs**
  * Main tabs in Member Page 

Implement new navigation logic into `Table` inside `TableRow`. And now we don't need to use **local state** in every component with `Table`. We just need to put the **length** of the list and **index** of **Row** inside `TableRow`.
And also add it into the `Pagination` component.

### Problems:
* At the moment didn't find a better solution for navigation in big lists. For now, everywhere used a pattern like in `VerticalMenu`
* Not sure about the style of _focused_ component. For now, it's just a blue border.
* Lots of rerenders on every navigating. We can see it clearly on **Dashboard** page (_greetings_ are changing)
* We can't focus on the **active tab** or **item** (in **tabs** and **sidebar**). Because the **navigation hook** saves just the first state of the **active item**. And when it changed it's still an **old** one inside the hook.

## Video
https://user-images.githubusercontent.com/89198006/147728836-12bef90b-cff7-4e77-8701-55bd4b729a8a.mov

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


hunt-hedvig:
if we go to a sidebar destination with mouse/trackpad clicks, then use the shortcut to go to another sidebar destination, the directory will point back to and reload the current page. Reproduction:
1. click dashboard with mouse/trackpad
2. press 'S' then arrow to select another sidebar item
3. -> re-entering the dashboard instead of going to the location at step 2

shared/@hedvig-ui/Table/table.tsx
```diff
@@ -166,20 +122,45 @@ export const TableRowStyled = styled.tr<{ active?: boolean; border?: boolean }>`
...
+            onResolve(index)
+          },
+          neighbors: {
+            up: index > 0 ? `Table Row ${index - 1}` : topElement,
+            down:
+              index < length - 1
+                ? `Table Row ${index + 1}`
+                : 'Table Pagination 0',
+          },
+        })
```
rasmusguterstam, line 160: I see what you mean with this being a bit problematic, as soon as the lists become pretty long and dynamic. This is OK for now though.

However, what happens if we have 2 tables on the same page? Won't the 'Table Row ${index}' be duplicated? 😅

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
```diff
@@ -285,7 +300,10 @@ export const useNavigation = () => {
       }
 
       return {
-        style: { border: '2px solid blue' },
+        style: {
+          // background: chroma(lightTheme.accent).brighten(0.5).alpha(0.5).hex(),
```
rasmusguterstam, line 304: I think your choice of color looks cool, so remove this comment!

src/portals/hope/features/claims/claim-templates/ClaimsTemplatesList.tsx
```diff
@@ -126,7 +154,8 @@ export const ClaimsTemplates: React.FC<ClaimsTemplatesProps> = ({
   )
 }
 
-interface TemplateCardProps {
+interface TemplateCardProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, 'onSelect'> {
```
rasmusguterstam, line 158: Nice!

src/portals/hope/features/claims/claims-list/filters/ClaimListFilters.tsx
```diff
@@ -191,136 +195,253 @@ export const ClaimListFilters: React.FC<ClaimListFiltersProps> = ({
...
+            },
+            neighbors: {
+              left: `Member Number ${range(numberMemberGroups)[0]}`,
+              up: index ? Object.keys(Market)[index - 1] : undefined,
+              down:
+                index < Object.keys(Market).length - 1
+                  ? Object.keys(Market)[index + 1]
+                  : undefined,
+              right: `Outcome Filter`,
+            },
```
rasmusguterstam, line 411:
```suggestion
          const markets = Object.keys(Market)
          const navigation = register(key, {
            resolve: () => {
              updateFilterHandler(UserSettingKey.MarketFilter, Market[key])
            },
            neighbors: {
              left: `Member Number ${range(numberMemberGroups)[0]}`,
              up: index ? markets[index - 1] : undefined,
              down:
                index < markets.length - 1
                  ? markets[index + 1]
                  : undefined,
              right: `Outcome Filter`,
            },
```

src/portals/hope/features/members-search/components/SearchForm.tsx
```diff
@@ -42,9 +45,23 @@ export const SearchForm: React.FC<SearchFieldProps> = ({
...
+  useEffect(() => {
+    if (searchFieldRef.current) {
+      if (cursor === 'Member Search Form') {
+        searchFieldRef.current.focus()
+        return
+      }
+
+      searchFieldRef.current.blur()
+    }
+  }, [cursor])
```
rasmusguterstam, line 64: What about the `autoFocus` prop instead of this? 😮

rasmusguterstam (approved):
Really good job ⭐ 

I think there's some stuff to resolve, but get that done then go ahead and merge. I think it's pretty clear that this implementation works well in some scenarios, and that it's very verbose in others - especially when the context gets very dynamic. 

Let's iterate and improve that another day, for now we'll consider navigaton done! ✅ 😄

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
maxplt: As I remember it's hard to see in some places in this color

shared/@hedvig-ui/Table/table.tsx
maxplt: Good catch! 
Maybe in this situation, we can add a prop like `id` or `name` and put it instead of `index` if it exists?

src/portals/hope/features/members-search/components/SearchForm.tsx
maxplt: Yeah, but here it doesn't work in this way 
Like it focused by color but not inside input 😰

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
maxplt: For example in `Tables` and in `Member Search Input`

src/portals/hope/features/members-search/components/SearchForm.tsx
maxplt: But we need _blur_ for example on key _Down_ and _focus_ again when we move to it from _table_ or _suggestions_. And `autoFocus` works only on the first render 😓

shared/@hedvig-ui/Table/table.tsx
rasmusguterstam: Yes! Or maybe even leverage the `key` prop? If it's even possible? Because it solves the same issue I guess, and every time we want to map our `TableRows` we'll have to specify it either way

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
rasmusguterstam: Ah alright! Well you decide which you think is better 😄

shared/@hedvig-ui/Table/table.tsx
maxplt: 😰

image

PR #549, opened 2022-01-11
Feature: Portals

rasmusguterstam:
# Jira Issue: [IP-216](https://hedvig.atlassian.net/browse/IP-216)

## What & Why
- Refactoring of directory and component structure on top-level to centralize things like authentication
- Portals are only visible and applicable if the role permits at least 2 portals. If only 1 is available, it's not possible to choose any other and the user won't even be exposed to the concept. The screenshots shown are from the perspective of a `Root` user. A SOS user will only see the SOS view, and IEX for example will only be able to see and use classic Hope.
- A global end-point `/portals` was added to facilitate debugging and eventually to be used instead of the current profile view (which might be replaced with just a button that routes to `/portals`). It's not accessible if only 1 portal is available.
- This feature is not really about SOS as much as it is about dissecting `hedgehog` into different, isolated internal products 

## Optional screenshots

https://user-images.githubusercontent.com/51323202/148962092-830678a5-36c7-4124-9b46-9f3be831dda5.mov

https://user-images.githubusercontent.com/51323202/148962274-9e9fa850-dd1c-4c22-9a9c-115bd0dad85b.mov

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -86,6 +86,7 @@ type Me {
   role: String!
   user: User!
   settings(filters: UserSettingsFilter): [UserSetting!]!
+  portal: String!
 }
 
 input UserSettingsFilter {
@@ -472,7 +473,7 @@ enum ContractStatus {
 
 type TermsAndConditions {
   commencementDate: LocalDate!
-  displayName: String!
+  displayName: String!,
   url: String!
 }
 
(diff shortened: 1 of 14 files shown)


fd-pliska (approved):
I have very little understanding of what's going on here, but it looks good! And, judging by the attached videos, it works, so let's go! :D

PR #563, opened 2022-01-24
Feature: Draft notes

maxplt:
# Jira Issue: [IP-233](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-233)

## What?
Make draft notes in claims with localStorage

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


rasmusguterstam (changes requested):
Alright so you've probably seen that we already have a hook called `useDraftMessages`, which this is a copy of but renamed. I would say that, instead of duplicating code, we should probably generalize to one hook (maybe `useDraft`) and extract it to `@hedvig-ui` or something. 

Because in general it doesn't really care about it being a `memberId` or `claimId` per se, but just a general identifier to refer a draft to.

rasmusguterstam:
On another note, we should add an event listener for storage to keep the draft in sync. This would enable having two windows in parallell and see the draft message being updated in one while writing it in the other 😎 

https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

shared/@hedvig-ui/hooks/use-draft.ts
```diff
@@ -0,0 +1,61 @@
+import { useInsecurePersistentState } from '@hedvig-ui/hooks/use-insecure-persistent-state'
+import { useEffect } from 'react'
+
+interface useDraftVariables {
```
rasmusguterstam, line 4:
```suggestion
interface UseDraftVariables {
```

shared/@hedvig-ui/hooks/use-draft.ts
```diff
@@ -0,0 +1,61 @@
...
+  id: string
+  section: string
```
rasmusguterstam, line 6: What is this?

shared/@hedvig-ui/hooks/use-draft.ts
```diff
@@ -0,0 +1,61 @@
...
+  expiry: number
+}
+
+export const useDraft = ({
+  id,
+  section,
+}: useDraftVariables): [string, (draft: string) => void] => {
+  const hour = 60 * 60 * 1000
+  const ttl = 24 * hour
+  const [drafts, setDrafts] = useInsecurePersistentState<Record<string, Draft>>(
```
rasmusguterstam, line 20: Is it really necessary to have a `section` now? I mean we'll only have `claimId` or `memberId`, so why not use the same?

src/portals/hope/features/conversations/chat/ConversationChat.tsx
```diff
@@ -49,7 +49,7 @@ export const ConversationChat: React.FC<{
   onBlur: () => void
   onResolve: () => void
 }> = ({ memberId, onFocus, onBlur, onResolve }) => {
-  const [draft, setDraft] = useDraftMessage({ memberId })
+  const [draft, setDraft] = useDraft({ id: memberId, section: 'conversation' })
```
rasmusguterstam, line 52: Maybe simplify the hook signature to just `(id: string)` instead of the object? This becomes so verbose. I know I created it originally, but since we're just doing a generic `id` now I'm thinking we might as well slim it down 😄

rasmusguterstam (changes requested):
Nice! Some things to resolve and some comments 😄

rasmusguterstam (approved):
Very clean 🌟

rasmusguterstam:
Don't forget to update `changelog.ts`!

PR #564, opened 2022-01-24
Chore: Don't allow implied any's

rasmusguterstam:
Just a lot of type fixes 🥴


src/portals/hope/features/dashboard/Greeting.tsx
```diff
@@ -12,9 +12,11 @@ const getDayPartGreeting = (name: string) => {
   } else if (hours >= 18 && hours <= 23) {
     return `Good evening ${name}!`
   }
+
+  return ''
```
vonElfvin, line 16: I guess something is better than nothing
```suggestion
  return `Hello ${name}`
```

vonElfvin (approved):
gg

PR #576, opened 2022-02-28
Feature: Hunter integration

rasmusguterstam:
## What?
- The things we demoed + some proper cleaning of the worst last-minute decisions code wise 🤓 
- Integration with Hunter
- Currently available at route `/search`, but will add way to reach it in a small upcoming PR

## Why?
- Since it's sort of hidden, this is just so we can get it out
- 

https://user-images.githubusercontent.com/51323202/155959269-40fd4623-2a8b-4f82-ae76-3b40e3acbfcd.mov


src/portals/hope/features/members-search/styles.tsx
```diff
@@ -10,22 +10,22 @@ const fadeIn = (max: number) =>
...
-  code: {
-    background: theme.backgroundTransparent,
-    padding: '1px 2px',
-    borderRadius: 1,
-  },
-  opacity: 0,
-  animation: `${fadeIn(0.3)} 1000ms forwards`,
-  animationDelay: '500ms',
-}))
+export const Instructions = styled.div`
```
maxplt, line 13: Nice 🤩
Don't understand why we can't do styles in this way everywhere

maxplt (approved):
Love it! Super clean 😍

src/portals/hope/features/members-search/styles.tsx
rasmusguterstam: We can, but it's usually so old code that people don't bother 😄

PR #577, opened 2022-02-28
make the Auth service the primary login provider on staging

bystam:
# Jira Issue: []

## What?
- Add the missing bits to use the auth service alongside gatekeeper
- Make the auth service default on staging

## Why?
- We want to roll it out on staging first.

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,6 @@ OAUTH_CLIENT_ID=<a uuid>
 OAUTH_CLIENT_SECRET=<very secure>
 API_URL=http://localhost:8443
 GATEKEEPER_HOST=<optional gatekeeper host>
-AUTH_HOST=<optional auth host>
 USE_STAGING_SPECIFIC_TOOLS=false
 USE_HELMET=false
 SWISH_PAYOUTS_ENABLED=false
--- a/.env.example
+++ b/.env.example
@@ -2,6 +2,7 @@ OAUTH_CLIENT_ID=<a uuid>
 OAUTH_CLIENT_SECRET=<very secure>
 API_URL=http://localhost:8443
 GATEKEEPER_HOST=<optional gatekeeper host>
+AUTH_HOST=<optional auth host>
 USE_STAGING_SPECIFIC_TOOLS=false
 USE_HELMET=false
 SWISH_PAYOUTS_ENABLED=false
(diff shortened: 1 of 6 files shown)


.env.example
```diff
@@ -2,6 +2,7 @@ OAUTH_CLIENT_ID=<a uuid>
 OAUTH_CLIENT_SECRET=<very secure>
 API_URL=http://localhost:8443
 GATEKEEPER_HOST=<optional gatekeeper host>
+AUTH_HOST=<optional auth host>
```
bystam, line 5: I guess this needs to be added somewhere?

src/portals/sos/pages/MainPage.tsx
```diff
@@ -28,21 +28,26 @@ const Footer = styled.div`
...
+        .hex()};
     margin-top: 4vh;
     transition: color 200ms;
 
     :hover {
-      color: ${({ theme }) => chroma(theme.foreground).brighten(2).hex()};
+      color: ${({ theme }) =>
+        chroma(theme.foreground)
+          .brighten(2)
+          .hex()};
```
bystam, line 42: This must've been auto-formatted through a git hook?

rasmusguterstam: Yeah I think there's a pre-commit prettify-something

src/entry.tsx
```diff
@@ -47,14 +47,12 @@ ReactDOM.render(
...
+              path="/login-provider"
               component={() => {
                 window.location.href = `${
-                  (
-                    window as Window &
-                      typeof global & { GATEKEEPER_HOST: string }
-                  ).GATEKEEPER_HOST
-                }/sso?redirect=${window.location.protocol}//${
+                  (window as Window & typeof global & { LOGIN_URL: string })
+                    .LOGIN_URL
```
rasmusguterstam, line 54: I'm guessing this?

```suggestion
                    .AUTH_HOST
```

rasmusguterstam: Aah nvm, saw more 🤓

.env.example
rasmusguterstam: Would it be controversial to just add the staging auth host here directly? This repo is public though, but it's not like we're exposing secrets 😄

rasmusguterstam (approved):

.env.example
bystam: Yeah, I will.

PR #578, opened 2022-02-28
Feature: Search tracking & useFeatureFlag

rasmusguterstam:
## What?
- Add `useFeatureFlag` hook to simplify feature flags
- Prepare tracking of search usage

## Why?
- Makes it one click away from adding the new search, will wait with that final part until Carl and I have solved updating the elasticsearch index dynamically 🥴


maxplt (approved):
Good one!

PR #589, opened 2022-03-10
Fix: Keyboard navigation

maxplt:
# Jira Issue: [IP-262](https://hedvig.atlassian.net/jira/software/projects/IP/boards/19?selectedIssue=IP-262)

## What?
Fix some issues and did some refactor:
*  `TopBar`: focus back on `notificationsButton` on modal closing by _**Escape**_
* Implement `use-navigation` inside `Questions` and make `Questions` more clear. 
* `Questions`: press _Enter_ to focus on `textarea`
  + Remove `use-hook-form` to come it true.
* Border bouncing. Fixed by _transparent_ border when no focus
* Fix issue when first listing item at lists not opening by Enter
* `Claims filters`: when hit _Enter_ stay focused at the filter group
* Re-rendring of Dashborad
  + By moving `Metrics` in other component

But I still have some stuff that I didn't resolve yet:
* Focus on the currently active or last visited item, but not the first child of the list/group (filters/sidebar items/tabs)
  + Not sure how to make it. Because **registration fields of navigation** don't update and always constant.
* Found other issue: on changing `claims filters` by navigation (_Enter_) they behave strangely. I think the reason is the same with previous issue

## Why?
For convenient keyboard navigation

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/portals/hope/features/claims/claim-details/ClaimFiles/FileUpload.tsx
```diff
@@ -35,7 +35,21 @@ export const FileUpload: React.FC<{
...
+    if (focus && buttonRef) {
+      buttonRef.current?.focus()
+      buttonRef.current?.scrollIntoView({
+        inline: 'center',
+        block: 'center',
+        behavior: 'smooth',
+      })
+    }
+  }, [focus])
+
```
rasmusguterstam, line 52: Sorry for being a pain here, but this is basically reintroducing the same pattern we completely reverted before. It's basically a question of component responsibility, where I see this as a _very_ slippery slope where a lot of different components from completely different features or parts of the application should explicitly respect the navigation feature by passing, reacting to and managing different focus states. 

That's why I tried playing around with centralizing this with the navigation context, where you basically "inject" with the context hook and are provided a nice interface (e.g. just spreading components) and things just work as expected. That's also why I decided to keep the `focus` state internal to the navigation context - as a developer, you should never care about what is currently being focused and not, and if you do, you should do that internally within the navigation context. I noticed that you exposed that state in a feature a while ago (i.e. being able to do something like `const { cursor } = useNavigation()`), and I let that slide to see if we were on to something, but the more I think about it I believe we should keep the `focus` / `cursor` state internally within the context and that it's a pretty good rule of thumb to keep it that way. 

This comment is not just for this piece of code, but basically the entire PR. Sorry that I can't give you an exact way to do this, but I can't help thinking that if we were to just move ahead with this, it will: (1) never be used (2) we'll slowly lose keyboard coverage over time since no one will maintain this and (3) developers will just be annoyed since they have to understand how it works, instead of just using a nice interface.

rasmusguterstam (changes requested):

src/portals/hope/features/claims/claim-details/ClaimInformation/ClaimInformation.tsx
```diff
@@ -301,19 +304,57 @@ export const ClaimInformation: React.FC<{
...
+          <ClaimOutcomeDropdown
+            claimId={claimId}
+            {...register('Claim Outcome', {
+              parent: 'Claim Card #2',
+              neighbors: {
+                up: 'Claim Status',
+                down: 'Date of Occurrence',
+              },
+            })}
+          />
```
rasmusguterstam, line 328: I like it and it's clean, but what about trying to navigate between these using tab index instead?

src/portals/hope/features/claims/claim-details/ClaimInformation/ClaimInformation.tsx
```diff
@@ -322,9 +363,22 @@ export const ClaimInformation: React.FC<{
...
+                    up: 'Contract for Claim',
+                    down: 'Employee Claim',
+                  },
+                },
+                {
+                  fontSize: '0.9rem',
+                },
+                {
+                  fontSize: '0.9rem',
+                },
```
rasmusguterstam, line 380: What 😅

src/portals/hope/features/claims/claim-details/ClaimInformation/ClaimInformation.tsx
```diff
@@ -336,11 +390,32 @@ export const ClaimInformation: React.FC<{
...
-          <ClaimEmployeeDropdown claimId={claimId} />
+          <ClaimEmployeeDropdown
+            claimId={claimId}
+            {...register('Employee Claim', {
+              parent: 'Claim Card #2',
+              neighbors: {
+                up:
+                  agreement && agreement.termsAndConditions?.url
+                    ? 'Terms And Conditions'
+                    : 'Contract for Claim',
```
rasmusguterstam, line 401: Not sure exactly how this would work, but what one could specify a list of strings instead and `useNavigation` tries them in that order?

So something like this, where it will move on to the next one if the other one isn't registered?

```typescript
up: ['Terms And Conditions', 'Contract for Claim']
```

rasmusguterstam: Thoughts?

src/portals/hope/features/claims/claim-templates/CreateFilterModal.tsx
```diff
@@ -56,16 +57,26 @@ export const CreateFilterModal: React.FC<CreateFilterProps> = ({
     onClose()
   }
 
+  const { register, focus } = useNavigation()
+
+  useEffect(() => {
+    focus('CreateFilterModal')
+
+    return () => focus('Add Template')
+  }, [])
```
rasmusguterstam, line 66: This is clean, but would it be possible to integrate similar logic within the `register` hook? The on-mount should be easy to fix, and the un-mount could maybe be just the parent? So when the component is un-mounted _or_ Escape is pressed, it will take you to `parent`? Thoughts?

src/portals/hope/features/claims/claims-list/filters/ClaimListFilters.tsx
```diff
@@ -369,23 +394,34 @@ export const ClaimListFilters: React.FC<ClaimListFiltersProps> = ({
...
                 : undefined,
               down:
                 index < range(numberMemberGroups).length - 1
                   ? `Member Number ${range(numberMemberGroups)[index + 1]}`
                   : undefined,
-              right: Object.keys(Market)[0],
+              right:
+                markets[index < markets.length ? index : markets.length - 1],
             },
           })
```
rasmusguterstam, line 426: In scenarios like these, it would be really cool if we could have something like `registerList` which given a list and a name will do this repetetive logic behind the scenes. maybe something like


```typescript

const { registerList } = useNavigation()

const { registerListItem } = registerList(
    range(numberMemberGroups),  // The list
    'MemberGroups',  // Namespace for the list items
    'MemberMarket',  // If we reach the bottom, where do we go?
    'SearchInput',    // If we reach the top, where do we go?
    (group) => console.log(group) // What do we do when we resolve the list item?
)

...

range(numberMemberGroups).map((filterNumber, index) => {
    return <SomeComponent {...registerListItem()} />
}


```

rasmusguterstam (changes requested):
Much better! Added some suggestions and some questions, would love to get your thoughts on all of it 😄

src/portals/hope/features/claims/claim-details/ClaimInformation/ClaimInformation.tsx
maxplt: Fully agree. Will try!

src/portals/hope/features/claims/claim-details/ClaimInformation/ClaimInformation.tsx
maxplt: Yeah, sure. At the moment purpose is to find a right way for items focusing.

src/portals/hope/features/claims/claim-templates/CreateFilterModal.tsx
maxplt: Nice point! I'm now trying to integrate this one inside `use-navigation` hook. No success at the moment :(

src/portals/hope/features/claims/claim-details/ClaimInformation/ClaimInformation.tsx
rasmusguterstam: @maxplt Any progress on this?

src/portals/hope/features/navigation/sidebar/VerticalMenu.tsx
```diff
@@ -259,6 +259,23 @@ export const VerticalMenu: React.FC = () => {
...
+    resolve: (item) => item.hotkeyHandler(),
+    focusCondition: (name) => {
+      const itemTitle = name.split(' - ')[1]
+
+      const route = MenuItemsList.find(
+        (menuItem) => menuItem.title === itemTitle,
+      )?.route
+
+      return route ? history.location.pathname.includes(route) : false
+    },
```
rasmusguterstam, line 276: What is this and why do we need it? 😮

maxplt: For the "focus on the currently active or last visited item"

maxplt: We need it for focusing on active item

src/portals/hope/features/claims/claim-details/ClaimInformation/ClaimInformation.tsx
maxplt: Made it in last commit. But at the moment it looks a little bit weird

src/portals/hope/features/claims/claim-details/ClaimPayments/components/ClaimPaymentForm.tsx
```diff
@@ -287,13 +303,31 @@ export const ClaimPaymentForm: React.FC<{
...
+            'Claim Payments Ex Gratia',
+            {
+              parent: 'Claim Card #6',
+              neighbors: {
+                up: 'Claim Payments Note',
+                down: 'Claim Payments Type',
+              },
+            },
+            { width: '8rem', marginBottom: '1.5rem' },
+            { width: '8rem', marginBottom: '1.5rem' },
```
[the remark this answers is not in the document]
maxplt: Yeah. But we need it to specify styles for navigation items, because some of them looks weird without these styles. And I can't find any better solution at the moment.

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
```diff
@@ -306,34 +382,100 @@ export const useNavigation = () => {
...
+          inline: 'center',
+          block: 'center',
+          behavior: 'smooth',
+        })
+      },
+      id: name,
+    }
+  }
+
+  function registerList<T>({
```
rasmusguterstam, line 424: How come `function` here??

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
```diff
@@ -306,34 +382,100 @@ export const useNavigation = () => {
...
+    focus,
+    resolve,
+    focusCondition,
+    withFocusCondition,
+    withFocus,
+    focusTarget,
+  }: {
+    list: T[]
+    name: string
+    nameField: keyof T
```
rasmusguterstam, line 437: 🌟

src/portals/hope/features/claims/claim-details/ClaimFiles/FileUpload.tsx
```diff
@@ -65,7 +68,7 @@ export const FileUpload: React.FC<{
       <FileUploadContainer>
         <Dropzone onDrop={handleDrop}>
           {({ getRootProps, getInputProps, isDragActive }) => (
-            <Button {...getRootProps()}>
+            <Button {...getRootProps()} {...register('Claim File Input', {})}>
```
rasmusguterstam, line 71: Do we need an empty json here? Shouldn't that just be default if we don't specify anything?

src/portals/hope/features/claims/claim-details/ClaimInformation/components/ClaimOutcomeDropdown.tsx
```diff
@@ -85,7 +85,7 @@ const useClaimOutcome = (claimId: string): UseClaimOutcomeResult => {
 
 export const ClaimOutcomeDropdown: React.FC<{
   claimId: string
-}> = ({ claimId }) => {
+}> = ({ claimId, ...props }) => {
```
rasmusguterstam, line 88: I guess this works, but maybe we should extend it with the `DropdownProps` as well to be clear? So something like

```typescript
React.FC<{ claimId: string } & Omit<DropdownProps, 'placeholder'>>
```

rasmusguterstam: For all of these dropdowns ofc

src/portals/hope/features/members-search/components/SearchForm.tsx
```diff
@@ -85,8 +72,10 @@ export const SearchForm = React.forwardRef<HTMLInputElement, SearchFieldProps>(
         <Group>
           <SearchInputGroup>
             <Input
-              style={{ borderRadius: '0.5rem' }}
-              {...register('Member Search Form', { autoFocus: true })}
+              {...register('Member Search Form', {
+                autoFocus: true,
+                withFocus: true,
```
rasmusguterstam, line 77: What's the difference? 🤓

src/portals/hope/features/questions/questions-list/QuestionGroupItem.tsx
```diff
@@ -26,46 +27,42 @@ const QuestionGroupWrapper = styled.div<{ isVisible: boolean }>`
...
+  HTMLDivElement,
+  QuestionGroupItemProps
+>(({ questionGroup, isFocused, ...props }, forwardRef) => {
   const [isVisible, setVisible] = React.useState(false)
 
-  const groupRef = useRef<HTMLDivElement>(null)
+  const internalRef = useRef<HTMLDivElement>(null)
 
-  React.useEffect(() => {
+  const ref = forwardRef ?? internalRef
```
rasmusguterstam, line 45: 🤝

src/portals/hope/pages/claims/ClaimDetailsPage.tsx
```diff
@@ -172,13 +190,67 @@ const ClaimDetailsPage: Page<
...
-            <Card span={3}>
+            <Card
+              span={3}
+              {...register(
+                'Claim Card #2',
+                {
+                  focus: Keys.Two,
+                  metaKey: 'ctrlKey',
+                  withFocus: true,
+                  focusTarget: 'Claim Status',
```
rasmusguterstam, line 222: Is this what happens when you press the focus key, in this case `ctrl + 2`? Because `withFocus` I guess is implied by `focusTarget` being defined, and also why isn't the `Claim Status` responsible for this logic instead of the `Claim Card #2`?

rasmusguterstam: Is it because you still want the card styled?

rasmusguterstam (changes requested):
Very close now, but some questions!

src/portals/hope/features/claims/claim-details/ClaimInformation/components/ClaimOutcomeDropdown.tsx
maxplt: I think I can remove this from changed dropdowns. Since we will use Tab navigation instead of keyboard.

src/portals/hope/features/members-search/components/SearchForm.tsx
maxplt: `autoFocus` - for do automatic **focus** via `useNavigation` on mount
`withFocus` (+ `focusTarget`) - for doing exactly **focus** to the `focusTarget` item (`.focus()`)

src/portals/hope/pages/claims/ClaimDetailsPage.tsx
maxplt: If there is no `focusTarget` but has `withFocus` - `true`, it'll be **focused** on the **current** item instead of the **target**.

maxplt: Used `Claim Card #2` for this purpose because we don't need to use keyboard navigation inside cards (we use tab navigation).
And yes, for some styles. You press `Ctrl + 2` and see, that you are focused on the Card. Then you press `Enter` and get focus on `Claim Status`

src/portals/hope/features/claims/claim-details/ClaimInformation/components/ClaimOutcomeDropdown.tsx
rasmusguterstam: Yeah good point!

src/portals/hope/features/members-search/components/SearchForm.tsx
rasmusguterstam: Alright maybe clarify by renaming to `withFocusTarget` or something?

src/portals/hope/pages/claims/ClaimDetailsPage.tsx
rasmusguterstam: Ah alright! Maybe split these up and have `focusTarget: string` and another, `focusOnElement` or something like that, so you know what to expect. If `focusTarget` is specified, it is implied that `withFocus` is true so that's just redundant

src/portals/hope/features/navigation/sidebar/VerticalMenu.tsx
```diff
@@ -246,6 +246,34 @@ export const VerticalMenu: React.FC = () => {
...
+
+      if (
+        history.location.pathname.includes('/claims') &&
+        route === routes.claims
+      ) {
+        return true
+      }
+
+      return route ? history.location.pathname.includes(route) : false
+    },
```
rasmusguterstam, line 274: Hm do we really need this? Why can't it just keep track of the list in the order we register them? Should be possible to generify this logic behind the scenes

src/portals/hope/features/questions/questions-list/FilteredQuestionGroups.tsx
```diff
@@ -33,44 +28,44 @@ export const FilteredQuestionGroups: React.FC<{
...
+            return (
+              <QuestionGroupItem
+                delay={`${index * 100}ms`}
+                key={questionGroup.id}
+                isFocused={questionGroup.memberId === focusedItem || false}
+                questionGroup={questionGroup}
+                {...navigation}
+              />
+            )
+          })}
```
rasmusguterstam, line 68: `registerList`?

src/portals/hope/features/user/notifications/NotificationsModal.tsx
```diff
@@ -41,12 +42,19 @@ export const NotificationsModal: React.FC<{
...
+    focus('NotificationsModal')
+
     markAllNotificationsAsRead({
       refetchQueries: ['GetMe'],
     })
+
+    return () => {
+      focus('NotificationsButton')
+    }
   }, [])
```
rasmusguterstam, line 58: ❌

rasmusguterstam: No explicit focus and/or leakage of `cursor` state (just remove them from the hook I guess)

src/portals/hope/pages/claims/ClaimDetailsPage.tsx
```diff
@@ -211,21 +300,74 @@ const ClaimDetailsPage: Page<
...
+                    withFocus: true,
+                    focusTarget: 'Claim Payments Payout',
+                  },
+                  {
+                    border: '2px solid blue',
+                  },
+                  {
+                    border: '2px solid transparent',
+                  },
+                )}
```
rasmusguterstam, line 342: Ah I see the point of `focusTarget` now! That's nice. I mentioned it in another comment, but `withFocus` is redundant here. And the double styles feels really weird, but I guess it's fine for now!

rasmusguterstam (changes requested):
😄

src/portals/hope/features/navigation/sidebar/VerticalMenu.tsx
maxplt: Behind the scenes, hook can't track the current situation. I didn't find a way of updating outside states. Like, inside hook we always have only the initial state.
This is the only way I found to recognize the changes.

src/portals/hope/features/members-search/components/SearchForm.tsx
maxplt: But it is not exactly only for the target. If an item has no **target**, it'll be **focused** on the current item (if `withFocus` is `true`)

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
```diff
@@ -134,63 +165,107 @@ export const NavigationProvider: React.FC<{ children: React.ReactNode }> = ({
...
           ? target.neighbors.up()
-          : target.neighbors.up
+          : typeof target.neighbors.up === 'object' &&
+            target.neighbors.up.find((item) => registry.current[item])
+          ? (target.neighbors.up.find(
+              (item) => registry.current[item],
+            ) as string)
+          : typeof target.neighbors.up !== 'object'
+          ? target.neighbors.up
+          : ''
```
rasmusguterstam, line 178: Hehe I see what you're doing, but simplify this into different lines and provide some explanatory names to make it easier to understand

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
```diff
@@ -134,63 +165,107 @@ export const NavigationProvider: React.FC<{ children: React.ReactNode }> = ({
...
           ? target.neighbors.left()
-          : target.neighbors.left
+          : typeof target.neighbors.left === 'object' &&
+            target.neighbors.left.find((item) => registry.current[item])
+          ? (target.neighbors.left.find(
+              (item) => registry.current[item],
+            ) as string)
+          : typeof target.neighbors.left !== 'object'
+          ? target.neighbors.left
+          : ''
```
rasmusguterstam, line 232: 🤪

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
```diff
@@ -134,63 +165,107 @@ export const NavigationProvider: React.FC<{ children: React.ReactNode }> = ({
...
           ? target.neighbors.right()
-          : target.neighbors.right
+          : typeof target.neighbors.right === 'object' &&
+            target.neighbors.right.find((item) => registry.current[item])
+          ? (target.neighbors.right.find(
+              (item) => registry.current[item],
+            ) as string)
+          : typeof target.neighbors.right !== 'object'
+          ? target.neighbors.right
+          : ''
```
rasmusguterstam, line 259: 🥴

shared/@hedvig-ui/hooks/navigation/use-navigation.tsx
```diff
@@ -306,34 +389,99 @@ export const useNavigation = () => {
...
+            down:
+              itemIndex < list.length - 1
+                ? `${name} - ${list[itemIndex + 1][nameField]}`
+                : undefined,
+          },
+          focusCondition,
+          withFocus,
+          focusTarget,
+          autoFocus: autoFocus ? list.indexOf(item) === 0 : false,
+        })
```
rasmusguterstam, line 478: Sweeet!! 🌟

rasmusguterstam: Skip for now, but makes it very easy to add a horizontal equivalent! 🤓

src/portals/hope/features/dashboard/MetricList.tsx
```diff
@@ -0,0 +1,279 @@
...
+                right:
+                  index < templateFilters.length - 1
+                    ? templateFilters[index + 1].name
+                    : 'Add Template',
+              },
+            },
+            {
+              background: chroma(lightTheme.accent).brighten(0.8).hex(),
+            },
+          )
```
rasmusguterstam, line 225: Or you know what, maybe it's actually worth adding the same thing as a vertical list but horizontally as well - would simplify this a lot, right?

src/portals/hope/features/navigation/sidebar/VerticalMenu.tsx
maxplt: @rasmusguterstam 
I mean that we can't do like this:
```
const navigation = register(item.title, {
   focus: index === 0 || pathname === item.title ? Keys.S : undefined,
   resolve: () => {
      tem.hotkeyHandler()
   }
   // ...
}
```
Because it saves only the path, that was on mounting. And it'll not be changed on redirections.
And in this way that provides `focusCondition` we achieve this.

src/portals/hope/features/dashboard/MetricList.tsx
```diff
@@ -0,0 +1,214 @@
...
+              {...registerItem(metric)}
+            />
+          ) : (
+            <AddMetricCard
+              {...registerItem(metric)}
+              tabIndex={0}
+              onClick={() => setCreateFilter(true)}
+            >
+              <Plus />
+              <span>Filtered Claim Template</span>
```
rasmusguterstam, line 202: Sweeet! 🌟

src/portals/hope/features/members-search/components/SearchForm.tsx
```diff
@@ -98,7 +87,7 @@ export const SearchForm = React.forwardRef<HTMLInputElement, SearchFieldProps>(
               }}
               icon={<SearchIcon muted={!query} />}
               placeholder="Looking for someone...?"
-              id="query"
+              // id="query"
```
rasmusguterstam, line 90: Remove if unused

src/portals/hope/features/navigation/topbar/TopBar.tsx
```diff
@@ -192,14 +192,16 @@ export const TopBar = () => {
...
+          style={
+            register('UsersOnlineButton', {
+              resolve: () => {
+                setShowUsers(true)
+              },
+              neighbors: {
+                left: 'NotificationsButton',
+              },
+            }).style
+          }
```
rasmusguterstam, line 204: Why do we have to do like this just here?

rasmusguterstam (approved):

PR #602, opened 2022-03-15
Put markets on User, require check-in and filter out markets

rasmusguterstam:
## What?
- Add `useMyMarkets` hook to add/remove/list the user's markets
- If no markets are set, require selection on check-in
- Market filters are still available as a regular filter, but will be restricted to your focus markets

https://user-images.githubusercontent.com/51323202/158392079-d935e873-c370-4c14-a98c-0aa9eb06dd9c.mov


src/portals/hope/pages/QuestionsPage.tsx
```diff
@@ -105,7 +115,12 @@ const QuestionsPage: Page = () => {
       </Spacing>
 
       <QuestionGroups
-        selectedFilters={selectedFilters}
+        selectedFilters={selectedFilters.filter((filter) => {
+          if (filter >= 6 && filter <= 8) {
+            return userMarkets.includes(intermediateMarketFilterMap[filter])
+          }
+          return true
+        })}
```
rasmusguterstam, line 123: This is really hacky, I know - but it comes from the user settings being stored as numbers, and hopefully we can kill this off pretty soon

PR #609, opened 2022-03-18
Fix: Lead window to renew access token

rasmusguterstam:
## What?
A component that generates a unique ID for the current window and registers itself as the lead window to be responsible for renewing the access token. This guarantees that at most one window will attempt to refresh when it expires. It is to be mounted top-level and, most importantly, outside `ApolloProvider` which instantiates the Apollo client.

## How it works
1. If it is the first mount of the window, register a unique ID for the window in session storage. The scope of Session Storage is per window, so this won't be shared.
2. Register the same ID in local storage to assign this window as "lead". If another window is opened, it will repeat the process but assign itself as "lead" instead.
3. When the Access Token expires, the window will first check whether it's the lead window. If it is, it will renew the access token. If not, it won't do anything.
4. Profit


vonElfvin (approved):
😃 🤝 😃

PR #615, opened 2022-03-25
Feature: Animations back

maxplt:
# Jira Issue: []

## What?
Made animations with **‘Framer Motion’** one more time. Since I accidentally merged to master Draft PR with it and then revert, I created a new one.
So, add some animations for `Modals`, `TasksPage`, `ChatPane`, `TopBar` and `Tabs`.

## Why?
Just for cosmetic improvements

## Video
https://user-images.githubusercontent.com/89198006/160056294-91f653a5-fb2e-4486-ae49-9c7f55a0a7e6.mov

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


vonElfvin (commented):
Really neat!

This could be a good tool in the toolbox, I think it looks good and makes the product feel more professional (we should make sure we don't over do it in the future though 🙈 )

I think the level you show in the video is good though!

`css` animations are pretty easy to add though, I'm not sure how this compares to that though? Maybe this makes it easier to add? I know we've done some `FadeIn` components and the like to do something similar to this before using `css` only.

Also, do we have to drill `visible`? 🤔

src/portals/hope/features/member/tabs/ChatPane.tsx
```diff
@@ -93,24 +89,30 @@ export const ChatPane: React.FC<{ memberId: string }> = ({ memberId }) => {
...
+          <ChatPanel memberId={memberId} />
+        </>
+      ) : (
+        <ChatHeader
+          visible={isVisible}
+          onResizeClick={onResizeClick}
+          isHinting={isHintingOption}
+        />
+      )}
+    </ChatContainer>
```
rasmusguterstam, line 115: Hm yeah as Elvin said, is this really necessary? 😮

rasmusguterstam: Since we're doing this check outside already. I guess its a matter of preference, but it feels like we're doing both here 😅

rasmusguterstam (approved):
Love it!! Only thing that maybe felt a bit overkill was the wiggly tabs on click and hover haha? 😄

PR #626, opened 2022-04-04
Update readme

fd-pliska:
## What?
- Update readme so that it states the version of Node needed to run the project smoothly


maxplt (approved):
Maybe add hints like 
`nvm install node@16.13.0`
or/and
`nvm use 16.13.0`

PR #641, opened 2022-04-14
Fix: questions resolve

maxplt:
# Jira Issue: []

## What?
Add question resolve in command line

## Why?
At the moment resolving only from input

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


rasmusguterstam (approved):
I assume the command line is blocked when the input is focused? So we don't fire twice 😅

PR #650, opened 2022-04-25
Fix: 'Left' templates

maxplt:
# Jira Issue: []

## What?
Fix issue when templates are staying even if you change the task

## Why?
-

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


rasmusguterstam (approved):

src/portals/hope/pages/tasks/TasksPage.tsx
```diff
@@ -270,6 +278,28 @@ const TasksPage: Page = () => {
...
+    if (
+      pickedLocale &&
+      formatLocale(pickedLocale as PickedLocale, true) ===
+        formatLocale(PickedLocale.EnSe, true)
+    ) {
+      changeLocaleDisplayed(memberId, true)
+    }
+
+    setLocale((pickedLocale || PickedLocale.SvSe) as PickedLocale)
+  }, [activeTask])
```
rasmusguterstam, line 301: Can we embed this in `useTemplateMessages`? Because the only input you have here is just `memberId` and `pickedLocale`, so would be sweet with a helper which does all this behind the scenes and only takes those two variables as input.

PR #657, opened 2022-04-28
Add button to cancel auto rescue

XSwifter:
# Jira Issue: [PAY-78]

## What?
- Add button to cancel auto rescue

## Why?
- 

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

[PAY-78]: https://hedvig.atlassian.net/browse/PAY-78?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/portals/hope/features/member/tabs/payments-tab/PaymentsTab.tsx
```diff
@@ -179,6 +182,33 @@ const MemberTransactionsTable: React.FC<{ memberId: string }> = ({
               <PaymentColumn title={transaction.status?.toString()}>
                 {transaction.status}
               </PaymentColumn>
+              <PaymentColumn title="Cancel Auto Rescue">
```
FredrikAreschoug, line 185: Can we some how se if this transaction is done through adyen? If so can we hide the button if it is not through adyen

FredrikAreschoug (commented):
Looks good added a comment

src/portals/hope/features/member/tabs/payments-tab/PaymentsTab.tsx
FredrikAreschoug: And also just show the button if it is in state `INITIATED`

XSwifter: Will try 😄

schema.graphql
```diff
@@ -240,6 +240,8 @@ type MutationType {
 
   auth_createApplication(applicationName: String!): AuthApplication!
   auth_generateCredentials(applicationId: ID!): AuthCredentialGenerationOutput!
+
+  cancelAutoRescue(transactionId: String): String
```
rasmusguterstam, line 244: Also, what is the string that we're returning? Since I'm guessing we're modifying the resource `Transaction`, I think it would be better if we also returned the mutated resource. Also, assume that it works (i.e. non-nullable), and let the clients handle the error if the mutation fails.

```suggestion
  cancelAutoRescue(transactionId: String!): Transaction!
```

src/portals/hope/features/member/tabs/payments-tab/PaymentsTab.tsx
```diff
@@ -179,6 +182,33 @@ const MemberTransactionsTable: React.FC<{ memberId: string }> = ({
...
+                        loading: 'Cancelling auto rescue...',
+                        success: ({ data: response }) => {
+                          if (response?.cancelAutoRescue === null) {
+                            return null
+                          }
+
+                          return 'Auto rescue successfully canceled!'
+                        },
+                        error: 'Could not cancel auto rescue',
+                      },
```
rasmusguterstam, line 205: If you return the mutated resource as mentioned in my other comment, you can simplify this one to

```suggestion
                      {
                        loading: 'Cancelling auto rescue',
                        success: 'Auto rescue successfully canceled',
                        error: 'Could not cancel auto rescue',
                      },
```

Also curious, which field is changed on a `Transaction` when the mutation succeeds? Because if the new resource is returned, the cache will be updated automatically 😀

src/portals/hope/features/member/tabs/payments-tab/PaymentsTab.tsx
```diff
@@ -179,6 +182,33 @@ const MemberTransactionsTable: React.FC<{ memberId: string }> = ({
...
+              <PaymentColumn title="Cancel Auto Rescue">
+                <Button
+                  onClick={(e) => {
+                    e.preventDefault()
+                    toast.promise(
+                      cancelAutoRescueMutation({
+                        variables: {
+                          transactionId: transaction.id!,
+                        },
+                      }),
```
rasmusguterstam, line 194:
```suggestion
                    e.preventDefault()
                    if (!transaction?.id) return
                    toast.promise(
                      cancelAutoRescueMutation({
                        variables: {
                          transactionId: transaction.id,
                        },
                      }),
```

src/portals/hope/features/member/tabs/payments-tab/graphql/cancel-auto-rescue.graphql
```diff
@@ -0,0 +1,3 @@
+mutation CancelAutoRescue($transactionId: String!) {
+    cancelAutoRescue(transactionId: $transactionId, )
+}
```
rasmusguterstam, line 3:
```suggestion
mutation CancelAutoRescue($transactionId: String!) {
    cancelAutoRescue(transactionId: $transactionId) {
      id
      amount {
            amount
            currency
      }
      timestamp
      type
      status
    }
}
```

schema.graphql
XSwifter: Actually we are not modifying the resource on the spot because an async flow is started. The resource is modified only when Adyen notifies us back that the rescue was cancelled successfully.
So here I thought I would just return the `transactionId` everything went well in calling Adyen, or `null` if something went wrong.

src/portals/hope/features/member/tabs/payments-tab/PaymentsTab.tsx
XSwifter: The status will be changed, but not when the mutation succeeds, because it starts an async flow that will eventually change the status when Adyen notifies back that the rescue cancellation succeeded. 😀

schema.graphql
rasmusguterstam: Ah I see! Is there some intermediate step on a `Transaction`, like a pending state or something while it's waiting? Or is it just identical until it has been canceled?

rasmusguterstam (approved):

schema.graphql
XSwifter: Unfortunately, it is identical.

rasmusguterstam: Alright then 😄

src/portals/hope/features/member/tabs/payments-tab/PaymentsTab.tsx
XSwifter: Added a condition to disable the button if the state is not `INITIATED`, but we have no way to see if this is an Adyen transaction. :(

FredrikAreschoug (approved):
Looks good to me

PR #658, opened 2022-04-28
IP-293 change the navbar ordering

falklinus:
# Jira Issue: [IP-293]

## What?
- Change position of user menu in navbar

## Why?
- More intuitive ordering

## Optional screenshots
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


maxplt:
Are you sure that putting **name** between **buttons** is a good idea?
For me personally, it looks a little bit unusual, but I'm not sure 🧐

falklinus:
> Are you sure that putting **name** between **buttons** is a good idea? For me personally, it looks a little bit unusual, but I'm not sure 🧐

What do you think of the new version with spacing between to create sections?

PR #664, opened 2022-05-04
Feature/restrict tpa access

falklinus:
# Jira Issue: [IP-282](https://hedvig.atlassian.net/browse/IP-282)

## What?
- Restricting claims and member search


src/portals/hope/features/config/constants.ts
```diff
@@ -200,3 +200,14 @@ export const TypeOfContractType: Record<TypeOfContract, InsuranceType> = {
...
+export const TypeOfContractByInsuranceType: Record<
+  InsuranceType,
+  TypeOfContract[]
+> = Object.entries(TypeOfContractType).reduce<
+  Record<InsuranceType, TypeOfContract[]>
+>((agg, item) => {
+  if (agg[item[1]]) agg[item[1]].push(item[0] as TypeOfContract)
+  else agg[item[1]] = [item[0]] as TypeOfContract[]
+  return agg
+}, {} as Record<InsuranceType, TypeOfContract[]>)
```
maxplt, line 213: Looks a little bit not obvious 😅
And I'm not sure if we should put methods/functions in the constants file. If I'm not mistaken it must consist only of constants. 🧐
Maybe put this function in some separate file?

src/portals/hope/features/search/components/SearchCategoryButtons.tsx
```diff
@@ -28,9 +29,21 @@ export const SearchCategoryButtons: React.FC<{
...
+  const hasSearchQuotesPermission = useHasPermission('searchQuotes')
+  const hasSearchCartsPermission = useHasPermission('searchCarts')
+
+  const filteredSearchCategories = searchCategories.filter((category) => {
+    if (category === 'members') return hasSearchMembersPermission
+    if (category === 'quotes') return hasSearchQuotesPermission
+    if (category === 'carts') return hasSearchCartsPermission
+
+    return true
+  })
```
maxplt, line 42: What if instead use `switch` with `true` **by default**? 🧐

maxplt (approved):
Looks good! 🌟

src/portals/hope/features/config/constants.ts
maxplt, line 213: It looks very different from other stuff in this file. 😅
Maybe make a function of getting these types and put it somewhere else? (since `constants.ts` is for constants) 🧐
What do you think?

src/portals/hope/features/search/components/SearchCategoryButtons.tsx
```diff
@@ -28,9 +29,21 @@ export const SearchCategoryButtons: React.FC<{
   category: SearchCategory
   onChange: (category: SearchCategory) => void
 }> = ({ category, onChange }) => {
+  const hasSearchMembersPermission = useHasPermission('searchMembers')
+  const hasSearchQuotesPermission = useHasPermission('searchQuotes')
+  const hasSearchCartsPermission = useHasPermission('searchCarts')
```
vonElfvin, line 34: No reason to do this up front? May as well run it in the return

src/portals/hope/features/navigation/sidebar/index.tsx
```diff
@@ -270,6 +271,18 @@ export const isExternalLinkAvailable = (path: string, markets: Market[]) => {
 }
 
 export const Sidebar = () => {
+  const hasToolsPermission = useHasPermission('navigateTools')
+  const hasInboxPermission = useHasPermission('navigateInbox')
+
+  const filteredMenuList = menuList.map((items) =>
+    items.filter((item) => {
+      if (item.name === 'tools') return hasToolsPermission
+      if (item.name === 'inbox') return hasInboxPermission
```
vonElfvin, line 280: Switch here as well IMO

src/portals/hope/features/search/components/SearchCategoryButtons.tsx
rasmusguterstam: Nope, these are hooks so we have to do them here

src/portals/hope/common/hooks/use-has-permission.ts
```diff
@@ -10,24 +10,44 @@ gql`
...
-  inbox: ['IEX', 'IEX_EXTENDED', 'ROOT'],
+const permissionMap: Record<RestrictedResource, string[]> = {
+  inbox: ['IEX', 'IEX_EXTENDED', 'ROOT', 'DEV'],
+  navigateInbox: ['IEX', 'IEX_EXTENDED', 'ROOT', 'DEV'],
+  navigateTools: ['IEX', 'IEX_EXTENDED', 'ROOT', 'DEV'],
+  searchCarMembers: ['CAR_TPA'],
+  searchQuotes: ['IEX', 'IEX_EXTENDED', 'ROOT', 'DEV'],
+  searchCarts: ['IEX', 'IEX_EXTENDED', 'ROOT', 'DEV'],
+  searchMembers: ['IEX', 'IEX_EXTENDED', 'ROOT', 'DEV'],
+  listCarClaims: ['CAR_TPA'],
```
vonElfvin, line 40: Something feels weird to me to have the word "car" as a explicit permission. If we're gonna need to do it like this for every time someone should work with a particular insurance it doesn't scale well. Maybe we can spend a few minutes discussing if we can make this in a more generic way

PR #673, opened 2022-05-16
Feature: Improve keyboard navigation

maxplt:
# Jira Issue: []

## What?
Make some kind of new keyboard navigation.
Attempt to use component with navigation instead of a hook.

### Main features: 
* It now provides a fresh `state` and correct updating of it inside navigation!
* And so, it is much easier to focus on current active item! You just need to specify the `focus` or `autoFocus` field! (In the previous version it doesn’t work as expected and has a `focusCondition` field which does not always work)
* Easier style specifying. You can specify basic navigation style by styled-components, or by style field. Also it provides style as function with focus parameter.
* Adding navigation items on **render**! In the previous version navigation in the component was initing anyway because it was a hook. Now you can specify a condition like when navigation need to be added/render. It resolves many issues.
* `Ref` and `focus`. So, now we don’t need to specify elements and add to them external `refs`, because now it specified in navigation component **by default**. And so, it has stable focusing and scrolling to view.
* And now it works correctly with mouse navigation. Like when you click anywhere, it doesn't break keyboard navigation! (or this issue doesn’t expect yet)

Pay attention to file `use-navigation`
Other files changes - just implementations of it.

## Why?
For better and stable keyboard navigation 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


maxplt:
# Video of usage

https://user-images.githubusercontent.com/89198006/168782064-1d25e15c-744c-41cc-ac3a-d02b9d97efc1.mov

vonElfvin (commented):
This looks like it works very well!

My one concern is if this is easy enough to use so that we'll continuously use keyboard navigation. The `useNavigation` hook was easy to tap into whereas this might be a bit clunky. What do you think yourself when using this?

maxplt:
> This looks like it works very well!
> 
> My one concern is if this is easy enough to use so that we'll continuously use keyboard navigation. The `useNavigation` hook was easy to tap into whereas this might be a bit clunky. What do you think yourself when using this?

@vonElfvin 
For me personally it's a little bit more easy to use and more clear what happens maybe. You don't need to call hook for register item (`const {register} = useNavigation()`), you just use wrapper with navigation. (Just like new Hotkeys for example). And for me, as I said, it is a little bit more convenient. And I think it's more flexible than hook.
Also it resolves some terrible issues, I wrote about it in above.
So, this is my point 🙂

vonElfvin (approved):
Looks really good in action! Let's try this out 🔥

PR #691, opened 2022-05-30
Feature: Member info

maxplt:
# Jira Issue: []

## What?
Make `Member Info` context for providing main information about member.

Also provide `preferred currency` on different mutations instead of hard-coded `'SEK'`.
But we have special hook `useContractMarketInfo` that already provides `preferredCurrency`. At the moment use new `Member Info` context for detect and use `prefferedCurrency`, but can replace it with `useContractMarketInfo`

## Why?
For more convenient access to main member information 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/portals/hope/features/member/index.tsx
```diff
@@ -79,6 +81,14 @@ export const MemberTabs: React.FC<{
       (member ? `${member?.firstName} ${member?.lastName}` : 'Loading...'),
   )
 
+  useEffect(() => {
+    if (member) {
+      updateMemberInfo(member)
+    }
+
+    return () => updateMemberInfo(null)
+  }, [member])
```
rasmusguterstam, line 90: I think it's problematic to designate the responsibility of updating a global context to one single component, especially when it's _very_ dependent on context.

rasmusguterstam: Yeah that would probably make more sense!

src/portals/hope/common/hooks/use-member-info.tsx
```diff
@@ -0,0 +1,46 @@
...
+      setMemberInfo(null)
+      return
+    }
+
+    setMemberInfo({
+      id: member.memberId,
+      currency: member.contractMarketInfo?.preferredCurrency || 'SEK',
+    })
+  }
+
```
rasmusguterstam, line 35: Would it make sense to have a specific query in here which dictates what data we're looking for? And instead just feed the `memberId` to the context

rasmusguterstam: I mean, there's nothing stopping me from accidentally feeding in any string here 🤪

rasmusguterstam: Also, it would give the context a better purpose; right now it's basically a global state management that you've decided to call "MemberInfoProvider". It's not providing any more value than e.g. Redux or Apollo Cache; but if we add a query, it actually serves a dedicated purpose

src/portals/hope/common/hooks/use-subclaim-reserve.ts
```diff
@@ -37,7 +39,7 @@ export const useSubclaimReserve = (subclaimId: string) => {
         subclaimId,
         amount: {
           amount: value,
-          currency: 'SEK',
+          currency: preferredCurrency || 'SEK',
```
vonElfvin, line 42: I'd rather explode than allow `'SEK'` as a fallback. Maybe do an early return of "Something went wrong fetching this member's currency" instead of the regular data

src/portals/hope/features/claims/claim-details/SubclaimPayments/components/PaymentConfirmationModal.tsx
```diff
@@ -84,11 +86,11 @@ export const PaymentConfirmationModal: React.FC<
       correctsPaymentId: selectedId ? selectedId : null,
       amount: {
         amount: +amount,
-        currency: 'SEK',
+        currency: preferredCurrency || 'SEK',
```
vonElfvin, line 89: Same here

src/portals/hope/features/claims/claim-details/SubclaimPayments/components/SubclaimPaymentForm.tsx
```diff
@@ -237,15 +241,15 @@ export const SubclaimPaymentForm: React.FC<{
           placeholder="Enter deductible"
           label="Deductible"
           step="any"
-          affix={{ content: 'SEK' }}
+          affix={{ content: preferredCurrency || 'SEK' }}
```
vonElfvin, line 244: You get the drill 🤪

vonElfvin (changes requested):
Great! Now the last thing is: "No currency = Explode" (or at least return a "Unable to do this" component of some sort)

vonElfvin (approved):

PR #699, opened 2022-06-09
remove the employees editor page

bystam:
# Jira Issue: []

## What?
- Remove the employee list editor

## Why?
- It is no longer in use, and replaced with Okta.

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


shared/@hedvig-ui/Row/row.tsx
```diff
@@ -0,0 +1,8 @@
+import styled from '@emotion/styled'
+
+export const Row = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+`
```
bystam, line 8: So there were a bunch of places where `Row` was imported from the employee editor, so I moved it here. Which looked like it was what we usually do? 🤷

vonElfvin (approved):
🔥 

Also, RIP @cpiehl1 's first project 🪦

bystam:
[Completely replace Gatekeeper with Auth service](https://trello.com/c/twggo3JO/16-completely-replace-gatekeeper-with-auth-service)

PR #724, opened 2022-06-16
Insurance based voucher campaigns

rasmusguterstam:
## What?
- Add multi-select dropdown to choose which insurance types a campaign is eligible for 

## Remarks
- Even though `GLOBAL` exists as a discount scope, we should probably stop using it and in the future be more explicit about which insurance types a campaign is for
- It really only makes sense for monthly percentage discounts

## Screenshots

image


rasmusguterstam:
[Implement discount scope when creating campaigns in Hope](https://trello.com/c/nfe4dPsW/32-implement-discount-scope-when-creating-campaigns-in-hope)

src/portals/hope/features/tools/campaign-codes/forms/MonthlyPercentageForm.tsx
```diff
@@ -167,6 +187,23 @@ export const MonthlyPercentageForm: React.FC = () => {
...
+        name="colors"
+        options={Object.values(InsuranceType).map((value) => ({
+          value,
+          label: convertEnumOrSentenceToTitle(value),
+        }))}
+        onChange={(newValues) =>
+          setFormData({
+            ...formData,
+            insuranceTypes: newValues.map(({ value }) => value),
+          })
```
falklinus, line 203:
```suggestion
          setFormData(prevState => ({
            ...prevState,
            insuranceTypes: newValues.map(({ value }) => value),
          }))
```

falklinus (approved):
Noice 👍

PR #728, opened 2022-06-20
Fix: Navigation clicks

maxplt:
# Jira Issue: []

## What?
Fix situations when clicks are unavailable when elements have navigation.

## Why?
To make clicking on elements with navigation available and to not specify click handler on _Navigation_ component every time.

## How?
Inside `onClick` event in _Navigation_ component do `onClick` event of **child**:
`(children as React.ReactElement).props?.onClick()`

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


falklinus (approved):
Neat 👍

PR #734, opened 2022-06-23
Use tasks query instead of user resolve to enable rollout of tasks

vonElfvin:
# Jira Issue: []

## What?
- Use tasks query instead of user resolve tasks

## Why?
- Enable rollout of tasks
- This is what's used under the hood, so now it's more explicit

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


falklinus (approved):
Fire

PR #748, opened 2022-06-30
Feature/pa 589/manual lead tool

Bjoernbom:
# Jira Issue: [PA-589](https://hedvig.atlassian.net/browse/PA-589)

## What?
- A tool to manually create a lead for all promises across all markets.

## Why?
- For us to be able to handle some of our leads for a non-technical person, we need to have a tool that can do so.

## Optional screenshots
image
image
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


src/portals/hope/features/tools/create-lead/forms/DanishHomeAccidentTravelFields.tsx
```diff
@@ -0,0 +1,101 @@
...
+    register,
+    formState: { errors },
+  } = useFormContext()
+  return (
+    <Fragment>
+      <Input
+        placeholder="Fuglevangsvej 11"
+        type="text"
+        label="Street"
+        error={errors['street']}
```
falklinus, line 17: Looking at the implementation of Input, I think the error prop is redundant if you use errors
```suggestion
```

falklinus: This also applies to the other fields of course.

falklinus (approved):
Nice job! Small PR 😅 I added a thought/suggestion regarding Input props.

src/portals/hope/features/tools/create-lead/forms/DanishHomeAccidentTravelFields.tsx
Bjoernbom: Okay I will look into it, thanks! 😃

Bjoernbom: Okay so if I removed the `error` prop but it seems like the Input field doesn't support only having `errors`, so I added support for that in the Input component. Please have a look @falklinus 😃

hugokallstrom (approved):
Would be really cool if we could dynamically generate the forms. Something for the future :) Otherwise LGTM!

Bjoernbom:
> 

@hugokallstrom Yea I agree and we've done that in the widget, but then you will have to specify each promise as a list of fields anyway somewhere. And in this scenario, each market has a different kind of validation, so yea I guess that's something for the future. However, I have an idea that there would be great to have a way of fetching all the fields, requirements, label, placeholder and other stuff for a specific promise, like PCMS. In that way implementations like this would be super easy. 😄

src/portals/hope/features/tools/create-lead/forms/DanishHomeAccidentTravelFields.tsx
falklinus: Nice 👍

PR #758, opened 2022-08-04
Fix: Inbox hotkey

maxplt:
# Jira Issue: []

## What?
When hitting `CMD + C` in **Inbox**, the **filter** is also triggered because of the `C` key involvement.
So, here is fix in this PR.

## Why?
For less conflicts between system built-in hotkeys and our custom hotkeys.

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


falklinus (commented):
Will this disable the cmd-c and cmd-v shortcuts everywhere?

maxplt:
> Will this disable the cmd-c and cmd-v shortcuts everywhere?

Yep. In my opinion it's reasonable, because they are binded by system. What do you think?

falklinus:
> > Will this disable the cmd-c and cmd-v shortcuts everywhere?
> 
> @falklinus Yep. In my opinion it's reasonable, because they are binded by system. What do you think?

@maxplt Yes I think it's good 👍

falklinus (approved):

PR #773, opened 2022-08-30
Refetch all tasks after changing preferences

vonElfvin:
# Jira Issue: []

## What?
- Refetch all tasks after changing preferences

## Why?
- So you get latest update on current tasks after preference change

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


falklinus (approved):
Coolio 🔥

PR #789, opened 2022-09-21
Finishing touches for claim type bindings

vonElfvin:
# Jira Issue: []

## What?
- Shorten user feedback for claim actions
- Don't attempt setting a default reserve until the type has been updated
- Make it easier to see which claim types that are bound using an opacity of `0.25` for those that aren't bound
- Correct `reserve` check when prompting user if a reserve is set or not
- Automatically reduce the `reserve` when making a payout

## Why?
- They annoyed me in 1 minutes, imagine working with them 40 hours a week
- So we can successfully set a reserve and not fail due to validation

## Optional screenshots
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


falklinus (commented):
Thoughts 🤔

PR #828, opened 2022-11-15
Feature/item models batch upload

falklinus:
# Jira Issue: [CLX-139](https://hedvig.atlassian.net/browse/CLX-139)

## What?
- Tabs system in item models tool
- Create: Enter data manually to create item model
- File upload: Upload item models from file
- Custom component for file upload
- Custom component for "select or input" (select that can turn into input)

## Why?
- Roadmap 👍 

## Optional screenshots
image
image

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


shared/@hedvig-ui/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export type FileUploadTableStatusType = Record<
+  number,
+  {
+    status: 'error' | 'success'
+    message: string
+  }
+>
+
+export const FileInput = ({
```
vonElfvin, line 28: Maybe this should have a more specific name for "excel" or "csv"

src/portals/hope/features/tools/item-models/UploadItemModels.tsx
```diff
@@ -0,0 +1,91 @@
...
+} from 'types/generated/graphql'
+import {
+  ItemModelBrand,
+  ItemModelType,
+} from 'portals/hope/features/config/constants'
+
+export const UploadItemModels = () => {
+  const [createItemModel] = useCreateItemModelMutation()
+  const [itemModels, setItemModels] = useState<CreateItemModelInput[]>([])
+  const [fileData, setFileData] = useState('')
```
vonElfvin, line 22: Since `FileInput` is very "excel heavy" maybe it can instead return an array of an array of strings so we don't need to `split('\n').split(',')` every time

vonElfvin: Might not work in practise since there are different delimiters

vonElfvin: I guess you could send that as input though `delimiter = ','` and it'll return the cells

vonElfvin (approved):
Really nice! We should probably update the backend to handle the scenario where some name already exists though. And simply change it's type/brand automatically. That way we can simply have an excel that is the "source of truth" that we send in over and over when we're making updates to it.

PR #829, opened 2022-11-16
use own local-storage hook instead of external

falklinus:
## What?
- Use own local storage hook for drafts, instead of external one

## Why?
- So that we can be in control of the logic, and use in dependency arrays

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


vonElfvin (approved):
Neat!

PR #834, opened 2022-11-21
Fix local storage hook and useDraft

falklinus:
## What?
- Fix the useDraft hook by using another localStorage hook

## Why?
- Issues with state not being shared among components

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


vonElfvin (approved):
Cool! If we want to it get even easier we can remove the `ttl`

vonElfvin (approved):

PR #838, opened 2022-11-23
Update/use backend types and brands for item models

falklinus:
## What?
- Fetch item model types and brands from backend
- Improve the "smart type and brand selector" of quick add model from input

## Why?
- Keep the source of thruth for types and brands in claims-service

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


src/portals/hope/features/tools/item-models/useItemModels.tsx
```diff
@@ -82,7 +118,7 @@ const useItemModelData = () => {
   }, [data?.itemModels])
 
   useEffect(() => {
-    fetchItemModels().catch(({ message }) => toast.error(message))
+    fetchItemModels().catch(null)
```
vonElfvin, line 121: Why catch?

vonElfvin (approved):
Maybe a bit too good UX for a simple CMS but now that we have it let's use it 😄

src/portals/hope/features/tools/item-models/useItemModels.tsx
falklinus: I got annoyed of the error that we're not handling the returned promise... seems like the toast.promise() is not considered handling it

PR #841, opened 2022-11-29
Disable polling in the Account query

schmee-hg:
This query is pretty heavy on account-service, and there is no reason to
poll this info cause it only changes when we run charges once a month or
if someone charges manually.


vonElfvin (approved):
This is very old code, I think we can fix the issue we tried to resolve initially with this polling by doing a proper cache-update when performing a mutation. But that requires changes both here and back-office...

PR #856, opened 2022-12-19
Suggest reference 

falklinus:
## What?
- When manual is selected, set reference to "Claim `memberId`" with "Claim hedvig" as fallback

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


src/portals/hope/features/claims/claim-details/SubclaimPaymentOrders/components/SubclaimPaymentOrderForm.tsx
```diff
@@ -155,9 +155,19 @@ const SubclaimPaymentOrderForm: React.FC<{
   }, [setValue, watch])
 
   useEffect(() => {
+    if (!watch('reference')) {
```
vonElfvin, line 158: Feels like we don't want it to be `hedvig`
```suggestion
    if (!watch('reference') && member?.memberId) {
```

vonElfvin (approved):
🔥

PR #857, opened 2022-12-19
Add a section which provides an easier text to copy-paste for mobile devs

StylianosGakis:
This is because sometimes we can't use the QR code but we do want to get the exchange token real quick.


vonElfvin (approved):
Izi

PR #874, opened 2023-01-20
Feature/auth group tool

vonElfvin:
# Jira Issue: [Security]

Note: These are already secured behind a role in the auth service so only a few people at Hedvig can set these even though they are accessible in Hope

## What?
- Add tool for adding Okta groups to `auth` and set their `permissions`

## Why?
- So we can actually start securing our `microservices`

- Use `gql` tag that doesn't work

## Optional screenshots
image

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


src/portals/hope/features/tools/auth-admin/components/AuthGroupAdmin.tsx
```diff
@@ -0,0 +1,179 @@
...
+    createAuthGroup,
+    setPermissions,
+  }
+}
+
+export const AuthGroupAdmin: React.FC = () => {
+  const { data } = useListAuthGroupsQuery()
+  const { createAuthGroup, setPermissions } = useAuthGroup()
+  const [groupName, setGroupName] = useState('')
+  if (!data?.auth_groups) return <Spinner />
```
falklinus, line 72: I guess this will work the same as { loading } from the query, so might as well use that for clarity. That way we also dont rely on it always returning an array

falklinus (approved):
Noice!

src/portals/hope/features/tools/auth-admin/components/AuthGroupAdmin.tsx
vonElfvin: Isn't `loading` also true if you refetch the query?

PR #884, opened 2023-01-31
Split member and memberclaims info into two requests

falklinus:
## What?
- Send two requests  for memberInformation to MemberPage

## Why?
- Claims information is restricted

- Make GraphQL return both data and errors

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


vonElfvin (approved):
Nice hack 🔥

PR #885, opened 2023-01-31
Update time "distance" for task createdAt

falklinus:
## What?
- Specify task time distance with two units

## Why?
- See more exact distance (eg. 1h 15min)

- Make api handle both future and past time

## Optional screenshots
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


src/portals/hope/features/tasks/list-items/TaskListItem.tsx
```diff
@@ -62,7 +64,7 @@ const Name = styled.div<{ available: boolean }>`
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
-  text- @media (max-width: 800px) {
+  text- @media(max-width: 800 px) {
```
vonElfvin, line 67: Sure about this one?

shared/@hedvig-ui/utils/date.ts
```diff
@@ -121,3 +126,63 @@ export const formatDistanceWithAccuracy = (
 
   return result.join(' ')
 }
+
+export const formatSmallDistanceWithAccuracy = (
```
vonElfvin, line 130: Feels like we're repeating ourself a lot here, especially since we already have `formatDistanceWithAccurracy` that does the top 3. But I cannot come up with a neat way to solve it now

vonElfvin (approved):
Coolio!

src/portals/hope/features/tasks/list-items/TaskListItem.tsx
falklinus: i will fix, think that is an error from before that the CI didn't catch

shared/@hedvig-ui/utils/date.ts
falklinus: Felt the same... 🤔

falklinus: @humledumle how would you go about avoiding something like "1 day 3 minutes". Imo that looks pretty weird 🤔

vonElfvin: I suggest we utilize `formatDistanceWithAccuracy` if it is more than 24 hours ago and hours + minutes part of this function when it's within a day. That simplifies a lot

vonElfvin: Hey @humledumle add me on Linkedin? 🙏  https://www.linkedin.com/in/elvin-granat-772461a8

PR #886, opened 2023-02-01
FIx impersonation clipboard bug for Android code

StylianosGakis:
## What?
- Android clipboard link didn't work as it copied some old structure where the token was saved inside `exchange-token` instead of the new `authorizationcode`

## Why?
- This resulted in the app not picking up the right token when passed into the deep link run configuration
image

- Don't have the right tools installed locally to format the code appropriately I think, I hope I didn't bork anything.
- Also I don't know how to run hedgehog locally so I simply did code changes so can't guarantee that it runs aside from knowing that the code I changed isn't that crazy 😂

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


vonElfvin (approved):
Nice DRY code

PR #908, opened 2023-02-15
Fix so we are logged out on a failed refresh

vonElfvin:
## What?
- Fix so we are logged out on a failed refresh

## Why?
- It should work

- Could maybe `catch` as well

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


src/apollo/client.ts
```diff
@@ -106,9 +106,12 @@ export const client = new ApolloClient({
...
 
-      refreshAccessToken().catch((e) => {
-        console.error('Failed to refresh access token', e)
-        toast.loading('Authentication failed')
+      refreshAccessToken().then((response) => {
+        if (!response || response.status < 400 || response.status >= 500) {
+          return
+        }
+        console.error('Failed to refresh access token')
+        toast.loading('Refreshing access token failed')
```
falklinus, line 114: I think you should also keep the catch, don't all statuses will go to `then`

falklinus (approved):

PR #919, opened 2023-02-22
IPC-1109: Implement Swedish legacy object insurance

Michael-Duivestein:
# Jira Issue: [IPC-1109]

## What?
- Implemented Swedish legacy object insurance.

## Why?
-

- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

[IPC-1109]: https://hedvig.atlassian.net/browse/IPC-1109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vonElfvin (approved):
If you want to show info about the object you need to do some stuff in the `AgreementInfo.tsx` file. It's pretty easy to understand the structure I think so you don't really need to be able to run it locally

src/portals/hope/features/config/constants.ts
```diff
@@ -144,7 +146,7 @@ export const QuoteProductTypeContractMap: Record<
     InsuranceType.NorwegianHouse,
     InsuranceType.DanishHouse,
   ],
-  [QuoteProductType.Object]: [],
+  [QuoteProductType.Object]: [InsuranceType.SwedishObjectlegacy],
```
mightyost, line 149: Should it not be added to the market const below on line 163 too?

mightyost (commented):
Urk, I thought HOPE/Hedgehog was totally agnostic to new types of insurances, i.e. I did not know you had to define a lot of stuff for this here :(!

Is it possible to make this 100% dynamic instead? No problems if new endpoints is needed in BO/UW to fetch the available insurances (type, typeOfContract, market, currency etc)!

vonElfvin:
> 

It works without you needing to do anything, you just need to do minor tweaks to improve the user experience if there is a new **type** of insurance.

vonElfvin:
> 

If you create an endpoint that has feature parity with this `constants.ts` file then yes we can do it dynamically. That's what we do (and more) for claims.

PR #920, opened 2023-02-23
Ability to skip sanction list on settle

vonElfvin:
# Jira Issue: [CLX-Büg]

## What?
- Ability to skip sanction list on settle

## Why?
- Payment orders are created even if sanction list is not skipped and the edit is a bit hard to find

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


falklinus (approved):
Noice

PR #936, opened 2023-03-14
Re-use cart search functionality in shop session search

cpiehl1:
# Jira Issue: [CLX-XXX]

## What?
- Generify the `QuoteCartResult` to be able to re-use it for displaying uw guideline breaches and overriding them for shop sessions

## Why?
- We want to reach feature parity with quote carts.

- Maybe we want to do something more tailored toward the way storefront works, but at the moment no one in IEX is familiar with all the new concepts so it is probably better to just stick to what they know.
- Maybe we could even have a single search that searches across shop sessions and quote carts?

## Optional screenshots
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


.env.example
```diff
@@ -7,4 +7,5 @@ SWISH_PAYOUTS_ENABLED=false
 USE_SECURE_COOKIES=false
 AUTH_HOST=https://auth.dev.hedvigit.com
 HEDVIG_ONBOARDING_URL=<a url>
+HEDVIG_STORE_URL=<a url>
```
cpiehl1, line 10: How do I actually inject this env variable in the staging and prod runtimes? 🙂

falklinus: https://dashboard.heroku.com/apps/hedvig-hope-staging/settings

falklinus (approved):

PR #960, opened 2023-04-12
Bazooka/contract picker

falklinus:
# Jira Issue: [CLX-522]

## What?
- Simplify contract picker props so we don't have to specify everything in the item itself where we want to use it
- Centralise the component

## Why?
- Allow changes more easily and improve dev ex
- DRY

- The @hedvigui component `Dropdown` is a bit hard to grasp and forces direct children to be options

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally

[CLX-522]: https://hedvig.atlassian.net/browse/CLX-522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/portals/hope/features/contracts/Trial.graphql
```diff
@@ -1,4 +1,4 @@
-fragment SubclaimTrial on Trial {
+fragment Trial on Trial {
```
vonElfvin, line 1: I'm guessing we can "get away with this" since we don't use fragments in the Contract queries in other parts of the system? 🤔

src/portals/hope/features/member/notes/MemberNotesModal.tsx
```diff
@@ -413,62 +413,15 @@ const FooterForm: React.FC<{
         (watch('activeTag') === MemberNoteType.ContractUpdate ||
           watch('activeTag') === MemberNoteType.HouseholdMember) && (
           <Flex style={{ marginBottom: 10, gap: '0.5rem' }}>
-            <Dropdown position="top" placeholder="None selected">
```
vonElfvin, line 416: Insane that these were copy pasted 🥲

vonElfvin (approved):
Beautiful! 😍

src/portals/hope/features/member/notes/MemberNotesModal.tsx
falklinus: haha yeah, super easy to miss when updating in one place 🫠

src/portals/hope/features/contracts/Trial.graphql
falklinus: Yeah i think it made more sense since we use the Contracts not only in claims, but also in member scope. It could be something like "AvailableTrialFragment" or something but I think this is good enough :)

PR #962, opened 2023-04-13
Bazooka/auth query

falklinus:
# Cleanup

## What?
- Get info on Me top level and pass to App

## Why?
- So we don't have to send two queries on every `Me` poll

- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


vonElfvin (approved):
Nice!

PR #968, opened 2023-04-24
Proactively refresh access token

falklinus:
# Jira Issue: [CLX-XXX]

## What?
- Track access token expiry and refresh when about 1min left

## Why?
- IEX being logged out when in the chat

## TODO
- Successfully refresh the token. It's broken right now

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/server/auth.ts
```diff
@@ -19,13 +18,19 @@ function setTokenCookies(
     path: '/',
     httpOnly: true,
     secure: config.useSecureCookies,
-    expires: addMinutes(new Date(), 29),
+    expires: undefined,
```
vonElfvin, line 21: Maybe we should have some "sanity" expiry?

src/server/request-enhancers.ts
```diff
@@ -31,6 +32,39 @@ export const setLoggerMiddleware: Middleware = async (ctx, next) => {
...
+
+export const setAuthMiddleWare: Middleware = async (ctx, next) => {
+  const accessTokenJwt = ctx.cookies.get('_hvg_at')
+  const accessToken = accessTokenJwt ? parseJwt(accessTokenJwt) : undefined
+  const secondsUntilExpiry = accessToken
+    ? getSecondsUntilExpiry(accessToken)
+    : undefined
+  const expiresSoon = secondsUntilExpiry ? secondsUntilExpiry < 60 : undefined
+  if (expiresSoon && canRefresh) {
+    setCanRefresh(false)
```
vonElfvin, line 61: I'm guessing through this we can only refresh one person at a time, which makes sense I guess since we have 60 sec buffer

src/server/request-enhancers.ts
```diff
@@ -31,6 +32,39 @@ export const setLoggerMiddleware: Middleware = async (ctx, next) => {
...
+  return secondsUntilExpiry
+}
+
+let canRefresh = true
+
+export function setCanRefresh(can: boolean) {
+  canRefresh = can
+}
+
+export const setAuthMiddleWare: Middleware = async (ctx, next) => {
```
vonElfvin, line 53: Maybe have name that is something more towards "attempt at proactive auth token refresh middleware"

vonElfvin (approved):
Nice! With this maybe we can simplify the refresh logic on the frontend later

falklinus:
Yes 👍 I think that's a good idea

PR #984, opened 2023-05-25
Remove whitelist

schmee-hg:
Was removed from member-service in https://github.com/HedvigInsurance/member-service/pull/780, and from back-office in https://github.com/HedvigInsurance/back-office/pull/841.


falklinus (approved):
Nice 👍

PR #986, opened 2023-05-30
Remove debt tab

schmee-hg:
I checked with IEX and we agreed to remove this on trial. If anyone
misses it within a week or so, we'll bring it back. Otherwise, less
code less problems!


falklinus (approved):
Nice 🔪

PR #1017, opened 2023-07-10
Feature/scheduled tasks and download files

vonElfvin:
## What?
- Fixes
  - Removed "Unassigning tasks from you" since it seems to bug when you go to "All" etc.
  - Fixed so calendar pickers start on Monday (Swedish+etc. weeks) instead of Sunday (US weeks)
  - Fixed so the "talking bubble" and "Fraudulent status" indicator are always circles in the Inbox (they don't get smushed together when a member has a long name)
  - Fixed so `onClickOutside` does not trigger when an element disappears on the click inside
  - Fixed so the cookie for the `at` and `rt` aligns with the actual expiry of the token, this should make it so you don't get logged out if you don't touch Hope in 24 hours due to the rt cookie being removed
  - Fixed so the Popover z-index can be specified, this due to Popovers added later would go higher in the hierarchy. Gives room for some configuration
  - Fixed optimistic response when changing task `area`

- Inbox
  - Updated design a bit (new icons for some quick options -> Area, Assing User, show assign reason on hover over user circle)
  
image

  - Fixed so quick options appear in the list more often for more people with more options
  - Added quick select "Assign to me" in list
  - Added "Schedule" ability on a task
    - Updated the design of "created at" to take into account when the "assignable from" is, it's green when it has not come into play and 
    -  Added a "Scheduled" tab in addition to "All" and "Unassigned" so the Unassigned list is not cluttered with tasks that have not come into play yet - considering adding another tab with "Scheduled to me" or similar so you can keep track of things assigned to you, but I don't know if assigning to oneself is a sought after behaviour just yet
    - Either as a quick option, where you keep the area it has
    - Or when selecting an area, you can then schedule it (a bit like a handover) - also renamed the options a bit so people don't abuse the fast one too much
    
image
image
   
 - Added ability to assign a particular person as well (with a quick select for yourself)
    
image

- Claim Files
image

  - Added ability to quick download a file (had to add a proxy in the koa server due to CORS issues)
  - Added ability to zip either all files or the selected folder and download the zip

## Why?
- A new Hope

- The caching of the s3 proxy is a bit weird, since the url is static and the request is a GET the client assumes it'll get the same response and uses the cached response. I'm using `no-cache` to force it to always call the request. Maybe one can figure out a way to simply change the URL. Maybe by not sending the presignedUrl as a header but a request param instead 🤷 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


package.json
```diff
@@ -136,6 +138,7 @@
     "@swc/jest": "0.2.24",
     "@testing-library/react": "12.1.5",
     "@types/chroma-js": "2.4.0",
+    "@types/file-saver": "^2",
```
vonElfvin, line 141: You can do this yourself but it requires like 7 lines of code. Might as well use a lightweight lib 🤷

shared/@hedvig-ui/Dropdown/dropdown.tsx
```diff
@@ -65,7 +65,7 @@ const OptionsList = styled.ul<{
   position: absolute;
   z-index: 10;
   width: 100%;
-  max-height: 500px;
+  max-height: 300px;
```
vonElfvin, line 68: Since it's searchable I figured it can be smaller and be prettier

shared/@hedvig-ui/OrbIndicator/orb-indicator.tsx
```diff
@@ -11,8 +11,8 @@ interface OrbIndicatorProps extends HTMLAttributes<HTMLDivElement> {
 }
 
 const OrbIndicatorStyled = styled.div<{ color: string; size?: string }>`
-  width: ${({ size }) => size};
-  height: ${({ size }) => size};
+  min-width: ${({ size }) => size};
+  min-height: ${({ size }) => size};
```
vonElfvin, line 15: Now it is a guaranteed to be a circle 🔴

shared/@hedvig-ui/SearchableDropdown/searchable-dropdown.tsx
```diff
@@ -94,12 +94,13 @@ export const SearchableDropdown: React.FC<
   Omit<
     CreatableSelectProps,
     'formatCreateLabel' | 'onCreateOption' | 'value'
-  > & { value?: string | number | null }
-> = ({ style, value, options, ...props }) => {
+  > & { value?: string | number | null; position?: MenuPlacement }
+> = ({ position = 'bottom', style, value, options, ...props }) => {
```
vonElfvin, line 98: Allows for searchable dropdown to grow upwards instead

shared/@hedvig-ui/TextDatePicker/index.tsx
```diff
@@ -98,6 +98,7 @@ const InlineDatePicker: React.FC<{
         {/*eslint-disable-next-line @typescript-eslint/ban-ts-comment*/}
         {/*  @ts-ignore*/}
         <DatePicker
+          calendarStartDay={1}
```
vonElfvin, line 101: Starts the weeks on Monday instead

shared/@hedvig-ui/hooks/use-click-outside.ts
```diff
@@ -20,8 +20,10 @@ export const useClickOutside = <T extends HTMLElement>(
       handler(event)
     }
 
-    document.addEventListener('mousedown', listener)
-    document.addEventListener('touchstart', listener)
+    // 👇 true to ensure it is run during "capturing" phase
+    // 👇 thus not incorrectly assuming elements that disappear are outside
+    document.addEventListener('mousedown', listener, true)
+    document.addEventListener('touchstart', listener, true)
```
vonElfvin, line 26: This failed on the ❌  to clear the searchable dropdown since it disappeared on click

src/portals/hope/features/tasks/hooks/use-create-task.ts
```diff
@@ -0,0 +1,78 @@
...
+    type: TaskResourceType,
+    assignableFromDay: ('TOMORROW' | 'MONDAY') | Date,
+    assignableTo: string | null,
+    note: string,
+  ) => {
+    const assignableFrom = (
+      assignableFromDay === 'TOMORROW'
+        ? process.env.NODE_ENV === 'production'
+          ? addHours(startOfTomorrow(), 8)
+          : addMinutes(new Date(), 1)
```
vonElfvin, line 44: "Tomorrow" is in 1 minute locally to test things

src/portals/hope/features/tasks/hooks/use-move-task.ts
```diff
@@ -14,17 +16,24 @@ export const useMoveTask = () => {
...
 
   const move = (taskId: string, email: string) =>
-    moveTask({
-      variables: { taskId, userEmail: email },
-      optimisticResponse: {
-        task_move: {
-          __typename: 'Task',
-          id: taskId,
-          assignedTo: email,
+    toast.promise(
```
vonElfvin, line 19: Moved the toasts to the root function to guarantee user feedback

falklinus (approved):
Chunky! Really nice additions and crisp ui/ux

PR #1018, opened 2023-07-10
Fix: Contract picker agreement

falklinus:
## What?
- show agreement according to dateOfOccurence in contract picker

## Why?
- information in contract picker is shown incorrectly 

- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/portals/hope/features/contracts/ContractItem.tsx
```diff
@@ -61,14 +61,17 @@ export const ContractItem: React.FC<{
...
 }> = ({ contract, preferredCurrency }) => {
-  const { getSubclaim, currentSubclaimId } = useClaim()
+  const { dateOfOccurrence } = useClaim()
 
-  const subclaimAgreement = currentSubclaimId
-    ? getSubclaim(currentSubclaimId)?.agreement
-    : undefined
   const title = convertEnumToTitle(contract.insuranceType)
 
-  const relevantAgreement = subclaimAgreement ?? contract?.currentAgreement
```
vonElfvin, line 71: Hmm, I think this is exactly the code for `subclaimAgreement` that is ran inside `back-office`

vonElfvin (approved):
Please see my comment, if I'm wrong let's try it!

src/portals/hope/features/contracts/ContractItem.tsx
falklinus: Yeah that's true. But it's only for the selected contract...

PR #1035, opened 2023-08-24
fixed darkmode

HubertClaesson:
# Jira Issue: [CLX-XXX]

## What?
- Faster animation on text when toggle darkmode 

Changed color on  active class in sidenav 

## Why?
- So that the text and icon doesn't disappear when clicked on in darkmode

Better ux   

- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


falklinus (approved):
Nice 👍

PR #1044, opened 2023-09-01
Update: View Reserve and Payments

falklinus:
## What?
1. View (not add) reserves and payments even when carrier is not present
2. Added a quick button to "remove" (set to 0) a reserve

## Why?
1. Sometimes we end up in a weird state where we don't have a carrier, but we have reserves and payments, the fix will allow IEX to view the reserve and payments, but not add new ones.
2. It can be used even when we don't have a carrier, so as to be able to change contract or date of occurrence after

- 

## Optional screenshots
image
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


vonElfvin (approved):
Nice!

PR #1046, opened 2023-09-04
Removed templates and latest changes from dashboardPage

HubertClaesson:
# Jira Issue: [CLX-XXX]

## What?
- Removed Templates and latest changes from DashboardPage.tsx

## Why?
- Templates was broken, it created two new templates when it should create one and few people used it  
-  Latest changes was not updated   
-  Makes room for new content

- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/portals/hope/pages/DashboardPage.tsx
```diff
@@ -1,108 +1,25 @@
...
-  useTitle,
-} from '@hedvig-ui'
-import { differenceInCalendarDays, format } from 'date-fns'
-import gql from 'graphql-tag'
+import { FadeIn, Spacing, useTitle } from '@hedvig-ui'
 import { Greeting } from 'portals/hope/features/dashboard/Greeting'
 import { useMe } from 'portals/hope/features/user/hooks/use-me'
 import React from 'react'
-import { DashboardNumbers } from 'types/generated/graphql'
-import { changelog } from '../../../changelog'
```
vonElfvin, line 18: Feels like you can clean up some more stuff that are only used by this file. For example I think this is unused now

vonElfvin (approved):
Nice! Please investigate my comment as well 🤝

src/portals/hope/pages/DashboardPage.tsx
```diff
@@ -1,108 +1,25 @@
...
-                      ` by ${change.authorGithubHandle}`}
-                  </MutedText>
-                </ChangeLogItem>
-              </FadeIn>
-            )
-          })}
-        </ChangeLogWrapper>
-      </Spacing>
+      <FadeIn></FadeIn>
+      <Spacing top="large"></Spacing>
```
falklinus, line 22: Shouldn't we keep the dashboard numbers on the page? 🤔 If not, FadeIn can be removed from here. Spacing can be removed anyway 🤔

HubertClaesson: What are the dashboard numbers?

vonElfvin: The numbers should still be around. Sorry I mixed up `Metrics` with `Changes` when I reviewed before. The `MetricsList` should be around. It's only the filters that should go

falklinus: Yes the numbers are rendered through the MetricsList component 👍

PR #1052, opened 2023-09-13
Fix: Adjusting reserves

falklinus:
## What?
- adjust correct reserve category when settling payment order
First update category reserve if there is one, then update the standard (Not specified) reserve

## Why?
-

- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


vonElfvin (approved):
Böm

PR #1053, opened 2023-09-13
Removed identity concept used in Norway.

mightyost:
Removed identity concept used in Norway.

Identity table in Member Service has not been populated since 2023-01-16 and Norway is sunsetting.


src/changelog.ts
falklinus, line 1: We don't display this anymore, will probably be removed entirely soon.

falklinus (approved):
Nice 👍

PR #1090, opened 2023-10-10
Remove: Test

falklinus:
## What?
- We don't have any tests: strip away unnecessary code/deps for it 

- If we want tests in the future, we can implement it the way we want 👍 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


vonElfvin (approved):
We do have some tests though? Simpler ones for string/date logics. But I guess they're not that important

falklinus:
@vonElfvin I don't think so 🤔 Can't find any, maybe they have been deleted earlier 🤷

PR #1091, opened 2023-10-10
Add CI script for main branch, bump eslint, prettier and yarn

falklinus:
## What?
- CI for main branch, lint & typecheck
- Semgrep is also run separately
- Bump eslint & prettier
- Fix linting & typecheck errors from bumping

## Why?
-

- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


.github/workflows/lint-typecheck.yml
```diff
@@ -0,0 +1,36 @@
...
+      - main
+
+jobs:
+  lint:
+    name: 'Lint & Typecheck'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
alebedev, line 18: v4 is available, can we use it?

.github/workflows/lint-typecheck.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Generate GraphQL types
+        run: yarn graphql:gen
+
+      - name: Lint
+        run: yarn lint
+
+      - name: Typecheck
```
alebedev, line 35: If we run semgrep here, will it be easier to maintain compared with separate actions?

alebedev (approved):
Nice!

.github/workflows/lint-typecheck.yml
alebedev: It's would be ideal to have single file in `.github/actions` - if we can do it

PR #1103, opened 2023-10-18
added bypass functionality

lucas-larsson:
## What?
- added bypass functionality to SAS EB insertion

## Why?
- This is to enable IEX to force add the EB number even if the member is not eligible. ( we do not know if the member is eligible unless we try to pay them)

-  Fix the SAS logic in `member-service`

### Without byPass
Screenshot 2023-10-18 at 17 24 23

### With byPass
Screenshot 2023-10-18 at 17 24 11

### Hover with redirect 
image

image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally

## Anything Else
I added the link to the notion page as a `const ` in the file.
 I was considering to add it as an `env` but I don't know if it is better practice because it is not directly a secret and I didn't want to make the deployment too messy with one more `env`, would like to hear input on this.


vonElfvin (approved):
Nice, please consider my comments though

PR #1120, opened 2023-10-26
Auth feature: HopeAuthGuard

falklinus:
## What?
- wrapper component for guarding content with auth permission

An application of leveraging that we have access to permissions on hope users.

- All hope pages are guarded with `backoffice, hope:read`
- Claims are guarded with `claims, claims:read`
- Search is guarded with `members, member-data:search`

## Optional screenshots
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/auth/components/HopeAuthGuard.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children,
+}: HopeAuthGuardProps) => {
+  const permissions = getAuthPermissions()
+
+  if (!permissions) {
+    return <NotAllowed />
+  }
+
+  const hasRequiredPermissions =
+    !requiredPermissions?.length ||
```
alebedev, line 20: No need to check length:
```
> [].every(() => false)
true
```

But you may still want to check for null / undefined

src/auth/permissions.ts
```diff
@@ -0,0 +1,78 @@
...
+      domain: AuthDomain.Members,
+      permission: 'member-data:search',
+    },
+  },
+  Storefront: {
+    read: { domain: AuthDomain.Storefront, permission: 'store-data:read' },
+  },
+} as const
+
+const parseJwt = (token: string) => {
```
alebedev, line 51: We're using `jwt-decode` lib in hedvig-com, can you look at it?  Parsing by hand is also fine if that's what you prefer

src/portals/hope/pages/HopePageRoutes.tsx
```diff
@@ -198,20 +199,64 @@ export const HopePageRoutes: React.FC = () => {
...
         <Route path="/questions" element={<TasksPage />} />
         <Route path="/inbox" element={<TasksPage />} />
-        <Route path="/search/:category" element={<SearchPage />} />
-        <Route path="/claims/list/:page?" element={<ClaimsListPage />} />
-        <Route path="/claims/:claimId" element={<ClaimPage />} />
+        <Route
+          path="/search/:category"
+          element={
+            <HopeAuthGuard
+              requiredPermissions={[
```
alebedev, line 206: Every usage I see has single element array, can you make this more convenient to use?  Something like `requires={HopeAuthPermissions.Members.searchMemberData}` while keeping array for places where we need to check multiple (if we have any)

alebedev (approved):
Left some comments, but looks good overall

src/auth/components/HopeAuthGuard.tsx
falklinus: if the list is empty, I want it to be true, do you think that will still be the outcome? 🤔 
And I figured checking ?.length also handles the case where it's undefined

src/auth/permissions.ts
falklinus: I can grab it if you think that's better! Just thought to not add a dependency for one function, but I guess that one is pretty nice!

src/auth/permissions.ts
```diff
@@ -0,0 +1,65 @@
...
+  const { accessToken } = authState()
+  if (!accessToken) return undefined
+
+  const parsedToken = jwt_decode<
+    JwtPayload & { permissions?: Record<string, string[]> }
+  >(accessToken)
+
+  if ('permissions' in parsedToken) {
+    return parsedToken.permissions
+  }
```
falklinus, line 62: @alebedev Is this similar to how you use it in hedvig.com?

src/portals/hope/pages/HopePageRoutes.tsx
falklinus: Nice suggestion! I updated it now

src/auth/components/HopeAuthGuard.tsx
falklinus: .every() returns true if the list is empty so yeah it was unnecessary 👍

PR #1125, opened 2023-11-13
Adjust reserve on settled orders

falklinus:
## What?
- Adjust reserve when settling multiple orders

## Why?
- Standard procedure

## Optional screenshots
image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/portals/hope/common/hooks/use-set-subclaim-reserve.ts
```diff
@@ -213,7 +213,110 @@ export const useSetSubclaimReserve = () => {
       )
     }
   }
+
+  const setReserveWithoutToast = async (
```
vonElfvin, line 217: Looks like a lot of code to simply be the same thing but no toast? Can't we have a boolean "showToast" on the other hook or something?

src/portals/hope/features/paymentOrders/hooks/use-payment-orders.tsx
```diff
@@ -233,6 +233,27 @@ export const usePaymentOrder = () => {
     }
   }
 
+  const adjustReserveWithoutToast = (
```
vonElfvin, line 236: Same thing here I guess

vonElfvin (commented):
Looks good UX wise, I think we should either refactor current function or rename these ones if all they do is not show a toast 🤝

src/portals/hope/common/hooks/use-set-subclaim-reserve.ts
falklinus: I wrote it in Improvements 🤪

vonElfvin: Don't you wanna fix that now that it is a bit fresh rather than later when you don't understand the code as much? 🤪

vonElfvin: Or are there hidden feature differences that are not as obvious in the name?

falklinus: Yeah I'm fixing, going with the boolean approach 👍

PR #1136, opened 2023-11-22
Restore behaviour to return to intended URL when redirecting to Okta

bystam:
# Notion Ticket: [CLX-XXX]

## What?
- Restore behaviour of remembering deep link when redirecting to Okta for re-logging in

## Why?
- It's nice!

- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/entry.tsx
```diff
@@ -66,26 +66,28 @@ const root = ReactDOM.createRoot(
 )
 
 const LoginProvider = () => {
-  useEffect(() => {
-    window.location.href = `${AUTH_BASE_URL}/login?redirect=${window.location.protocol}//${window.location.host}/login/callback`
-  }, [])
+  const returnTo = window.sessionStorage.getItem('hvg:login-return-to')
+  const redirect = `${window.location.protocol}//${window.location.host}/login/callback?return-to=${returnTo}`
+  window.location.href = `${AUTH_BASE_URL}/login?redirect=${encodeURIComponent(
```
alebedev, line 71: Does it work as expected?  When you run side effects in React render code, it can run undetermined number of times.  Hence `useEffect` in original version

src/entry.tsx
```diff
@@ -66,26 +66,28 @@ const root = ReactDOM.createRoot(
...
+  const [returnTo] = useSearchParamsState('return-to', '/')
 
   saveAuthTokens({
     access_token: accessToken,
     refresh_token: refreshToken,
     expires_in: parseInt(expiresIn),
   })
 
-  return <Navigate to={previousLocation ?? '/'} />
+  window.location.href = returnTo
```
alebedev, line 89: Same here

src/entry.tsx
falklinus: maybe try putting a console log to see if it runs multiple times. I scouted when he implemented and at least the functionality was working

bystam: I mean it also worked with the `useEffect`, and I think we should just re-instate it if we risk having problems without it. We made a very basic test, it could be that there were lurking bugs that we didn't find.

bystam: PR updated

alebedev (approved):
Looks good

falklinus (approved):
🔥

PR #1148, opened 2023-12-20
Chat Image definition

falklinus:


src/portals/hope/features/member/messages/components/Message.tsx
```diff
@@ -106,8 +106,16 @@ const Image = styled.img`
...
 
-const isImage = (text?: string) => {
-  return text?.toLowerCase()?.match(/\.(jpeg|jpg|gif|png|heic|webp)$/) != null
+const isImageContent = (content: MessageContentType) => {
+  const isImage = (text?: string) => {
+    return text?.toLowerCase()?.match(/\.(jpeg|jpg|gif|png|heic|webp)$/) != null
+  }
+
+  return (
+    content.mimeType?.includes('image') ||
```
vonElfvin, line 115: Doesn't this simply do the trick?

falklinus: maybe, was unsure if it would support old stuff 🤷

PR #1149, opened 2023-12-21
Feature: Extended Leave / Vacation mode

falklinus:
## What?
- Possibility to set Vacation status

## Why?
- This will "pause distribution" for the user which will affect the tasks with `assignableTo`, distributing them to checked in users instead

- I think the core concept and UX is acceptable but naming/implementation could be improved, a bit nonDry

## Optional screenshots

https://github.com/HedvigInsurance/hedgehog/assets/33982429/14b3a17f-209a-4d13-a864-c5b5d0b43a64

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


lucas-larsson (approved):
Looks great, Love the vacay mode. 

I thought just a bit about the copy in place, "No members will be assigned to you" is it members or tasks or what is it called??

Other than that. looks great

falklinus:
> Looks great, Love the vacay mode.
> 
> I thought just a bit about the copy in place, "No members will be assigned to you" is it members or tasks or what is it called??
> 
> Other than that. looks great

It's tasks, but IEX doesn't really use that lingo 👀

src/portals/hope/pages/tasks/TasksPage.tsx
```diff
@@ -1006,7 +1006,11 @@ const TasksPageInner: React.FC = () => {
                       }
                     >
                       <FilterBarItem
-                        onClick={pauseTaskDistribution}
+                        onClick={() =>
+                          pauseTaskDistribution(
+                            new Date(Date.now() + 10 * 60000),
```
bystam, line 1011: Worth putting it into a var like? :) 
```js
const tenMinutes = new Date(Date.now() + 10 * 60000)
```

bystam (approved):
Looks great :) one small suggestion

src/portals/hope/pages/tasks/TasksPage.tsx
falklinus: 👍

PR #1160, opened 2024-01-10
Ability to add informative label to guarded content

falklinus:
## Why?
- Inform "super-users" that something is usually not available

## Optional screenshots
image


fredriklagerblad (approved):
⭐

